PR #1, opened 2019-12-08
Subgraph Composition

Jannis:
This RFC proposes a new feature called _subgraph composition_ to allow subgraphs to reference and extend types from other subgraphs, either by subgraph name or by subgraph ID.

### DIFF
--- a/rfcs/0001-subgraph-composition.md
+++ b/rfcs/0001-subgraph-composition.md
@@ -1,4 +1,4 @@
-# RFC-0001: Subgraph composition
+# RFC-0001: Subgraph Composition
 
 <dl>
   <dt>Author</dt>
--- a/rfcs/0001-subgraph-composition.md
+++ b/rfcs/0001-subgraph-composition.md
@@ -185,10 +185,11 @@ At that point, failure to find an indexer for a subgraph that types were
 imported from will result in a query error.
 
 Until the network is reality, we are dealing with individual Graph nodes and
-querying subgraphs where foreign entities are not also indexed on the same node
-should be handled with more tolerance. This RFC proposes that entity references
-to a subgraph that is not available should simply resolve to `null` values in
-query results.
+querying subgraphs where imported entity types are not also indexed on the same
+node should be handled with more tolerance. This RFC proposes that entity
+reference fields that refer to imported types are converted to being optional in
+the generated API schema. If the subgraph that the type is imported from is not
+available on a node, such fields should resolve to `null`.
 
 ## Compatibility
 
--- a/rfcs/0001-subgraph-composition.md
+++ b/rfcs/0001-subgraph-composition.md
@@ -30,7 +30,7 @@ subgraph boundaries.
 
 The high-level goal of subgraph composition is to be able to compose subgraph
 schemas and data hierarchically. Imagine umbrella subgraphs that combine all the
-data from a domain (e.g. DeFi, job markets, music) under one unified, coherent
+data from a domain (e.g. DeFi, job markets, music) through one unified, coherent
 API. This could allow reuse and governance at different levels and go all the
 way to the top, fulfilling the vision of _the_ Graph.
 
@@ -44,15 +44,15 @@ enables several use cases:
    
 Subgraph composition is needed to avoid duplicated work, both in terms of
 developing subgraphs as well as indexing them. It is an essential part of the
-overal vision behind The Graph, as it allows to combine isolated subgraphs into
+overall vision behind The Graph, as it allows to combine isolated subgraphs into
 a complete, connected graph of the (decentralized) world's data.
 
 Subgraph developers will benefit from the ability to reference data from other
 subgraphs, saving them development time and enabling richer data models. dApp
 developers will be able to leverage this to build more compelling applications.
 Node operators will benefit from subgraph composition by having better insight
-into which subgraphs will be queried together, allowing them to make more
-informed decisions about which subgraphs to index.
+into which subgraphs are queried together, allowing them to make more informed
+decisions about which subgraphs to index.
 
 ## Urgency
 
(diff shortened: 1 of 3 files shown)


rfcs/0001-subgraph-composition.md
```diff
@@ -0,0 +1,233 @@
...
+subgraph.
+
+### Subgraph availability
+
+In the decentralized network, queries will be split and routed through the
+network based on what indexers are available and which subgraphs they index.
+At that point, failure to find an indexer for a subgraph that types were
+imported from will result in a query error.
+
+Until the network is reality, we are dealing with individual Graph nodes and
```
ghost: What about in the case where the foreign type is not nullable in the GraphQL schema?

rfcs/0001-subgraph-composition.md
```diff
@@ -0,0 +1,233 @@
...
+There are other ways to compose subgraph schemas using GraphQL technologies such
+as [schema
+stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching/)
+or [Apollo
+Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
+However, schema stitching is being deprecated and Apollo Federation requires a
+centralized server to serve a combined GraphQL API.
+
+Another reason not to use these is that GraphQL will only be _one_ of several
+query languages supported in the future. Composition therefor has to be
```
ghost: therefore*

rfcs/0001-subgraph-composition.md
```diff
@@ -0,0 +1,233 @@
...
+}
+```
+
+This makes queries like the following possible, where you can go "back" from
+accounts and contracts to proposal entities, despite the Ethereum `Address` type
+originally being defined in the `ethereum/mainnet` subgraph.
+
+```graphql
+{
+  accounts {
```
leoyvens: I don't follow this example, this queries `accounts` but I don't see an `Account` type anywhere.

rfcs/0001-subgraph-composition.md
leoyvens: This same open question exists in the design for having fields in a same entity come from different chains and networks within a same subgraph, the entities need to be merged by id and a mandatory field may be missing.

IMO tolerance is the right guiding principle, and a way to be tolerant here but not violate GraphQL rules is to omit entities which don't have all their mandatory fields after merging from the result.

rfcs/0001-subgraph-composition.md
```diff
@@ -0,0 +1,233 @@
...
+However, schema stitching is being deprecated and Apollo Federation requires a
+centralized server to serve a combined GraphQL API.
+
+Another reason not to use these is that GraphQL will only be _one_ of several
+query languages supported in the future. Composition therefor has to be
+implemented in a query-language-agnostic way.
+
+## Open Questions
+
+- An open question is whether it should be possible to extend foreign
```
leoyvens: What about implementing interfaces for foreign types?

rfcs/0001-subgraph-composition.md
Jannis: Oh, yeah, the query is outdated. This should `addresses`. I'll update it.

rfcs/0001-subgraph-composition.md
Jannis: Are we talking about
```
type SomeEntity @entity {
  foo: ForeignType!
}
```
where the subgraph from which `ForeignType` was imported is not available and the `foo` value would be `null`?

I don't think we can exclude entities that exist but are not broken from query results. IMHO what should happen in the above case is that the GraphQL query is satisfied but there is an error for the branch of the query that involves the entity. We should compare to the error behavior other GraphQL servers (e.g. ApolloServer) have when a non-nullable field in a response is not set.

rfcs/0001-subgraph-composition.md
Jannis: It's going to be fun to mix local and foreign concrete types that implement the same interface (local or foreign). It sounds feasible though.

rfcs/0001-subgraph-composition.md
leoyvens: The graphql spec has discussion on this exact case https://graphql.github.io/graphql-spec/draft/#sec-Errors.Error-result-format.

> If the field which experienced an error was declared as Non-Null, the null result will bubble up to the next nullable field.

This is similar to my suggestion, but the entity can't just vanish, it should be replaced by `null`.

If we include errors with data in our protocol, we need to consider what that will look like for the read interface and attestations.

rfcs/0001-subgraph-composition.md
leoyvens: So this is more of a future feature than an open question to be tackled in this RFC?

rfcs/0001-subgraph-composition.md
Jannis: @leoyvens Yep, that's a good point. I think we _do_ want to include errors like this in the response, so they will have to be part of the read interface / attestations as well. The query engine should either pick a node that has both/all subgraphs involved, or it will split the query up at the subgraph boundaries (which I'm less sure about in the composition case than in regular queries), but even then, the data the query engine has may be out of date or the index node may remove one of the subgraphs involved before it processes the query.

Everything but an easy situation _but_ for now I'll include some details about this paticular case with the desired error behavior to the RFC.

rfcs/0001-subgraph-composition.md
Jannis: I think we should tackle it. Will think about this a bit more.

Jannis: Done, added a section about interfaces.

Jannis:
@graphprotocol/developers Rebased. Anything we still need to update here?

leoyvens (approved):

yanivtal:
This looks perfect. You rock 🙌

PR #2, opened 2019-12-10
GraphQL query prefetching

lutter:


SUMMARY.md
```diff
@@ -7,5 +7,6 @@
   - [Rejected RFCs](./rfcs/rejected.md)
 - [Engineering Plans](./engineering-plans/index.md)
   - [Approved Plans](./engineering-plans/approved.md)
+    - [SQL query combination (prefetch)](./engineering-plans/0001-sql-query-combination.org)
```
Jannis: Add a `PLAN-0001:` prefix to the link caption and let's remove the `prefetch`.

engineering-plans/0001-sql-query-combination.org
```diff
@@ -0,0 +1,397 @@
...
+  <dd><a href="https://github.com/graphprotocol/rfcs/pull/2">https://github.com/graphprotocol/rfcs/pull/2</a></dd>
+
+  <dt>Date of submission</dt>
+  <dd>2019-11-27</dd>
+
+  <dt>Date of approval</dt>
+  <dd>2019-12-10</dd>
+
+  <dt>Approved by</dt>
+  <dd>Jannis, Leo</dd>
```
Jannis: Let's do full names here.

engineering-plans/0001-sql-query-combination.org
```diff
@@ -0,0 +1,397 @@
...
+</dl>
+
+This is not really a plan as it was written and discussed before we adopted
```
Jannis: Sadly, in contrast to GitHub, mdBook doesn't support Org Mode, so this will have to be converted to Markdown (let's drop the `<a>` anchors from the output though that Org Mode exports often include.

engineering-plans/0001-sql-query-combination.org
```diff
@@ -0,0 +1,397 @@
...
+|------+--------------+-----------------+-------------+----------------------------+------|
+|    1 | TRUE         | TRUE            | TRUE        | child.parents ∋ parent.id  | A    |
+|    2 | FALSE        | TRUE            | TRUE        | child.parents ∋ parent.id  | A    |
+|    3 | TRUE         | TRUE            | FALSE       | child.parent = parent.id   | B    |
+|    4 | FALSE        | TRUE            | FALSE       | child.parent = parent.id   | B    |
+|    5 | TRUE         | FALSE           | TRUE        | child.id ∈ parent.children | C    |
+|    6 | TRUE         | FALSE           | FALSE       | child.id ∈ parent.children | C    |
+|    7 | FALSE        | FALSE           | TRUE        | child.id = parent.child    | D    |
+|    8 | FALSE        | FALSE           | FALSE       | child.id = parent.child    | D    |
+
```
Jannis: For formatting this, we can use https://github.github.com/gfm/#tables-extension- in mdBook.

engineering-plans/approved.md
```diff
@@ -1,3 +1,3 @@
 # Approved Engineering Plans
 
-- No engineering plans have been approved yet.
+- [PLAN-0001: SQL Query Combination (Prefetch)](./0001-sql-query-combination.org)
```
Jannis: Drop ` (Prefetch)` from the title.

Jannis (changes requested):

Jannis:
@lutter I'm also ok renaming the entire feature from `SQL query combination` to `Query prefetching` or something.

lutter:
Renamed 'SQL query combination' to 'GraphQL query prefetching', and addressed all the other comments above.

Jannis (approved):

engineering-plans/0001-graphql-query-prefetching.md
```diff
@@ -0,0 +1,397 @@
...
+<dd><a href="<https://github.com/graphprotocol/rfcs/pull/2>"><https://github.com/graphprotocol/rfcs/pull/2></a></dd>
+
+<dt>Date of submission</dt>
+<dd>2019-11-27</dd>
+
+<dt>Date of approval</dt>
+<dd>2019-12-10</dd>
+
+  <dt>Approved by</dt>
+  <dd>Jannis Pohlmann, Leo Yvens</dd>
```
Jannis, line 20: I think @leoyvens name is Leo(nardo) Yvens Schwarzstein. 😉 Leo: Would you prefer to have the complete name in these RFCs or just a part of it?

engineering-plans/0001-graphql-query-prefetching.md
```diff
@@ -0,0 +1,397 @@
...
+
+
+## Implementation Details for prefetch queries
+
+
+### Goal
+
+For a GraphQL query of the form
+
+    query {
```
Jannis: Let's do code blocks with

    ```graphql
    ```
and
    ```sql
    ```
rather than indentation. That way we get syntax highlighting on GitHub and in mdBook.

Jannis (changes requested):

engineering-plans/0001-graphql-query-prefetching.md
```diff
@@ -0,0 +1,397 @@
...
+
+<dl>
+  <dt>Author</dt>
+  <dd>David Lutterkort</dd>
+
+<dt>Implements</dt>
+<dd>No RFC - no user visible changes</dd>
+
+<dt>Engineering Plan pull request</dt>
+<dd><a href="<https://github.com/graphprotocol/rfcs/pull/2>"><https://github.com/graphprotocol/rfcs/pull/2></a></dd>
```
Jannis: This links is broken ever so slightly. Should be
```
<a href="https://github.com/graphprotocol/rfcs/pull/2">https://github.com/graphprotocol/rfcs/pull/2</a>
```

Jannis (changes requested):

engineering-plans/0001-graphql-query-prefetching.md
leoyvens: I think the only ones who would remember how to spell 'Schwarzstein' are either me (most of the time) and those who speak german 😄

Leo Yvens is fine since that's also my handle.

leoyvens:
The windowing is still over my head. It looks like the query with windowing doesn't use `first` and `offset`, could the lack of `first` be a performance issue?

I also wonder what this would look like in a v2 API where `first` and `skip` are gone, the windowing would be needed to get the right order regardless?

engineering-plans/0001-graphql-query-prefetching.md
lutter: I wasn't sure what to put for you, but glad I guessed right. Not that I'd have a problem with saying Schwarzstein: Schwarzstein, Schwarzstein, Schwarzstein (one day you have to tell me how that's pronounced in Portuguese) ;)

engineering-plans/0001-graphql-query-prefetching.md
lutter: Arrgh, I had that in org-mode, but Emacs helpfully stripped it out

engineering-plans/0001-graphql-query-prefetching.md
lutter: Too much copy pasta. Fixed

lutter:
> The windowing is still over my head. It looks like the query with windowing doesn't use `first` and `offset`, could the lack of `first` be a performance issue?

Windowing splits the result set into separate bins/windows according to the criterion in `partition by`; we partition by `parent_id`, so the result set (what comes out of the `from` and `where` clauses) gets split into pieces, one for each distinct `parent_id`. Each of the windows is then sorted according to the `order by` in the window clause, and the `rank()` reports where in that sorted list each result row falls; we put that as the `pos` column into the result set.

We then take that result set and only take the entries whose `pos` falls into the range described by the `first` and `skip` arguments, so that at the end we have a result set with the children for each `parent_id` within that range (the first child _for each parent_ will have `pos == 1` etc.)

You can think of the traditional `limit` and `offset` in SQL queries as viewing the result set as one big window. Overall, the windowed queries are a bit slower than a query with just `limit / offset`, but those don't let us express what we need, which is 'get n children for each parent, according to this order`; and this is still much faster than running a query for each parent separately.

> I also wonder what this would look like in a v2 API where `first` and `skip` are gone, the windowing would be needed to get the right order regardless?

Yes, and I assume that we'd only do away with `skip`; so rather than 'give me the entities with rank n up to rank n+m' we'd say 'give me the m entities after the entity with this id', in both cases according to a fixed order. The queries would look similar to what you get today when `skip == 0`, with an additional clause to start at the right point, probably something like `sort_field > :last_value and id > :last_id` (for ascending order, same with `<` for descending order) and we would encode `(:last_value, :last_id)` into the query response as the token to get the next page. We could also just do something like `pos > :last_pos`; which way to go depends on which style of query will be faster in the end.

lutter:
Addressed the remaining comments.

engineering-plans/0001-graphql-query-prefetching.md
leoyvens: @lutter In Portuguese it's pronounced Schwarzenegger.

leoyvens:
@lutter Cool, I understand it better now. It's good that we'll have a simplified query for when windowing is unecessary. I'd be interested to see some quick experiments measuring the performance gain in doing the simplification that allows using `offset`/`first` instead of the windowing, when this is all implemented.

lutter:
I have the speedup measurements for [this issue](https://github.com/graphprotocol/graph-node/issues/1081) from before I added windowing to the prefetch code (those queries happened to produce the correct result without windowing)

Here is the data for the scenario in that issue with and without windowing (all times in ms):

| first | windowed | limit+offset | windowed / 1k | limit+offset / 1k |
|-------|----------|--------------|---------------|-------------------|
|    50 |      170 |           87 |          3400 |              1740 |
|   100 |      210 |          127 |          2100 |              1270 |
|   200 |      320 |          196 |          1600 |               980 |
|   500 |      640 |          437 |          1280 |               874 |
|  1000 |     1400 |          844 |          1400 |               844 |
|  2000 |     2500 |         1893 |          1250 |             946.5 |

It's not an entirely valid comparison, but points in the direction of windowing affecting query performance quite a bit; but in either case we see considerable speedup over the current GraphQL execution which takes about 15000ms for 1k entities.

engineering-plans/0001-graphql-query-prefetching.md
Jannis: Hahahaha

engineering-plans/0001-graphql-query-prefetching.md
Jannis: Yeah, it does that, sadly.

Jannis (approved):

Jannis:
@lutter Approved! Note: I've enabled only _merge_ commits in this repository to have more explicit and visible merge points for approved rfcs/plans.

leoyvens:
@lutter ah nice we already have the numbers. Yes that perf difference is not the most fortunate but it's perfectly acceptable, especially comparing to the current situation.

leoyvens (approved):

lutter:
Rebased to latest master

PR #4, opened 2019-12-13
Ethereum Tracing Cache RFC

That3Percent:


rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
+# RFC-0000: Template
```
Jannis: This would be `RFC-0002: Ethereum Tracing Cache`

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+<dl>
+  <dt>Author</dt>
+  <dd>Zac Burns</dd>
+
+  <dt>RFC pull request</dt>
+  <dd><a href="URL">URL</a></dd>
```
Jannis: This will always be a bit funny. After opening an RFC PR, the RFC itself has to be amended to replace `URL` with the PR URL.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+  <dt>Date of approval</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Approved by</dt>
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Summary
+
+This RFC proposes the creation of a local etherium tracing cache to speed up indexing of subgraphs which use callHandler.
```
Jannis: `etherium` -> `Ethereum`

Jannis: Instead of `callHandler`, I'd say `block and call handlers`, since I believe both are affected. If we wanted to reference the `callHandler` field it would be good to put them in backticks to mark them as inline code.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+
+If a subgraph has callHandlers specified, it is necessary to extract calls from a trace of each block that a graph node indexes. It is expensive to acquire and process traces from etherium nodes, both in money and time.
```
Jannis: callHandlers -> `block and call handlers`

Instead of `specified` I'd say `If a subgraph uses ...`.

Jannis: Our convention for `graph node` has been to either capitalize it as `Graph Node` or, if the text is closer to the code, use `graph-node` to refer to the executable. In this document, I'd use `Graph Node`.

Jannis: `etherium` -> `Ethereum`

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+A production graph node may index multiple subgraphs and repeatedly access the same traces many times - especially for recent blocks. Having that data on hand in a pre-processed and easily filtered format may substantially reduce the cost of acquiring and converting trace data into triggers relevant to a subgraph.
+
+## Urgency
+
+None
+
+## Terminology
+
+None.
```
Jannis: We're introducing at least one new term here for the feature itself: `Ethereum tracing cache`.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Detailed Design
+
+The design comes in 2 parts. The caching hierarchy, and the easily filtered data format.
+
+### Cache Hierarchy
+In `etherium_adapter.rs` is a function `traces`, which unconditionally makes a call to eth.web3. A hierarchical cache would be inserted here with 3 layers - In-Memory, On-Disk, and Fetch.
```
Jannis: This to me is an implementation detail that could go into the engineering plan. If we drop the mention of `ethereum_adapter.rs` here and just say the cache is designed to consist of three layers, then we're good.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### On-Disk Layer
+The In-Memory layer is not expected to remain in cache for very long in an effort to reduce memory consumption. An on-disk layer helps to prevent more calls to the fetch layer than is absolutely necessary.
+
+This on-disk layer would be implemented as a binary file store with a cache eviction policy over postgres. It is expected that the most recent blocks would be indexed the most frequently, so the scoring mechanism for the cache would probably be some mix between LRU, block recency, and block finality.
+
+#### Fetch layer
+The existing code in the `traces` function would be renamed to `fetch_traces` and would be the bottom layer of the cache. At first, this would exist like it does today. In the future this may be augmented to either delegate to fetching from a parity API or to calculate traces on demand using a local etherium node, depending on which resources are currently the bottleneck.
```
Jannis: My suggestion would be for these three layers to be modelled after the `eth_call` cache that is already built into Graph Node:

- The `Fetch Layer` would indeed be identical to today, where we simply make `trace_filter` or `trace_block` requests (depending on the situation) in `EthereumAdapter`.

- In addition, we introduce a new `EthereumTrac{e,ing}Cache` component much akin to the already existing `EthereumCallCache`. This component would store traces in a Postgres table and would "serve" traces either from memory of from the database. I don't think we need to specify whether this is as `rc::Weak`, `Arc` or something else in the RFC.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Futures re-use
+In order to prevent concurrent loads of data from one cache layer to another (Eg: multiple concurrent calls to the underlying eth.web3 API with the same parameters) the hierarchical cache would have a mechanism to delegate multiple tasks to the same future with a NonReady state.
```
Jannis: I'm not sure I get this part; it feels like an implementation detail as well.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+* Processable as a linear stream with minimal cache misses, ideally not requiring the entire file to be loaded at once.
+* Move redundant data outside of the Trace (eg: Block Number)
+
+JSON is the format the data is delivered in today. JSON is neither compressed nor fast: https://eng.uber.com/trip-data-squeeze/. JSON does not have any of the properties of our ideal format. The entirety of a JSON file needs to be parsed whether or not all of it is actually used. JSON's self-description and flexibility come at a large cost for data which has a schema. As much of the data in a trace is binary data or data which does not have an established representation in JSON, additional overheads are incurred in encoding representations of traces in the primitives that JSON supports.
+
+Flatbuffers seems like a good initial candidate here which makes reasonable tradeoffs between skip/zero-copy and compression ratios. Significant gains over Flatbuffers are possible with a custom format, but this is not likely a good place to spend time in the short term. 
+
+This is subject to change if it is decided that filtering/indexing should happen within postgres.
```
Jannis: I believe @leoyvens had some suggestions on storing only parts of each block's traces in the cache. Perhaps he can elaborate.

Regarding the size of traces: I've pulled traces for a fairly large block (~260 transactions) and they are 201kb large. Multiply that by ~9,000,000 blocks and you end up with ~1.8TB. So storing _all_ traces would require a lot of disk space and I/O. 

Each trace by itself does not have a crazy overhead:
```json
{
      "action": {
        "callType": "call",
        "from": "0x0681d8db095565fe8a346fa0277bffde9c0edbbf",
        "gas": "0x2d4b0",
        "input": "0xa9059cbb0000000000000000000000003f706d0da87ceef53ebf701a47fd38d5e4f1ed6700000000000000000000000000000000000000000000000000000000abd7a95e",
        "to": "0xdac17f958d2ee523a2206206994597c13d831ec7",
        "value": "0x0"
      },
      "blockHash": "0x030bf7a9e55f0178aa28c700c9346a3af9d3a7c513ab23f6a46d5b0b3496f0db",
      "blockNumber": 9115243,
      "result": {
        "gasUsed": "0x8729",
        "output": "0x"
      },
      "subtraces": 0,
      "traceAddress": [],
      "transactionHash": "0x15d6d3cad7cd2faa3b68330bd47672e569458b62ff46fe07dcb9dedf8b807b9e",
      "transactionPosition": 48,
      "type": "call"
    },
```
@olibearo can probably tell which parts we need here, I suspect we need most of these fields. What makes traces large is mainly the `action > input` and `action > output` fields; especially the inputs can contain the entire bytecode of a contract. I think it's also worth looking at the different trace types to decide what a good format to store them would be. I've seen `call`, `crate` and `reward` traces so far.

Developing our own binary file format for these feels like overkill. I'd rather see if we can narrow down what data we want to store and how we can fit that into a Postgres schema.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Drawbacks and Risks
+
+In some cases, a graph node may not frequently get a cache hit when requesting a trace. If cache hits are relatively rare, the additional cost of pre-processing the data and caching it would only add to the overhead that exists today. It is currently up to a combination of the implementation details of block_stream and the set of subgraphs being indexed as to whether this strategy would be a win for a particular graph node. It may be necessary to monitor statistics about the behavior of the caching layer to keep tabs on whether this change continues to be a win in a production environment.
+
+To mitigate this risk, it may be worth frontloading work on the cache hierarchy to gather statistics about block trace access patterns and how frequently cache hits are expected for existing graph nodes.
+
+## Alternatives
+
+A general purpose etherium cache has been suggested. It is likely that much of the work here would be able to be repurposed for such a cache.
```
Jannis: `etherium` -> `Ethereum` 😉

Jannis: We could restructure the existing `EthereumCallCache` to be a more general `EthereumCache` that also covers traces. This is a decision we can make when writing the engineering plan for the feature though; whether we do it or not shouldn't matter at all in terms of effort, since the surface area of both caches is narrow.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Open Questions
+
+What, if any, ordering constraints exist on triggers across multiple contract addresses?
```
Jannis: Ordering matters only within a subgraph. What we currently guarantee in terms of ordering is this (from https://thegraph.com/docs/define-a-subgraph#the-subgraph-manifest):

> The triggers for a data source within a block are ordered using the following process:
>
> 1. Event and call triggers are first ordered by transaction index within the block.
> 2. Event and call triggers with in the same transaction are ordered using a convention: event triggers first then call triggers, each type respecting the order they are defined in the manifest.
> 3. Block triggers are run after event and call triggers, in the order they are defined in the manifest.

Jannis (changes requested):

rfcs/0002-tracing-cache.md
leoyvens: Having the block as the cache key is the simplest and also guarantees that we only need to get the traces from Ethereum once for each block, but then if any trace in a block is needed all traces in it will be stored, which will bloat disk size with traces that are potentially never used. To balance this we could use a more fine grained key for the cache, for example `(block, signature, contract)` since the call triggers in a subgraph will always specify a signature and usually specify an address.

rfcs/0002-tracing-cache.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Detailed Design
+
+The design comes in 2 parts. The caching hierarchy, and the easily filtered data format.
+
+### Cache Hierarchy
+In `etherium_adapter.rs` is a function `traces`, which unconditionally makes a call to eth.web3. A hierarchical cache would be inserted here with 3 layers - In-Memory, On-Disk, and Fetch.
+
+#### In-Memory Layer
+At the top level of the cache is the in-memory cache. For a graph node which is indexing multiple subgraphs in parallel, it may be likely that they require the latest blocks simultaneously. By storing the buffer containing immutable trace data for a block in a rc::Weak, a low-risk, low-cost cache is introduced. The most recent trace, and all currently in-use traces would be maintained.
```
leoyvens: The in-memory cache is clever, but the gain from it is less clear, so I'd suggest leaving it as a future extension.

rfcs/0002-tracing-cache.md
That3Percent: This is trying to state that when multiple cache misses occur the work should not be performed twice concurrently. It is an implementation detail though, and I'll move it to the engineering plan.

rfcs/0002-tracing-cache.md
That3Percent: > Developing our own binary file format for these feels like overkill

> ... 201kb large. Multiply that by ~9,000,000 blocks and you end up with ~1.8TB

Those statements seem to conflict, unless our goal is to only store a very small percentage of blocks. (More on this in a later comment)

> What makes traces large is mainly the action > input and action > output fields;

I think what makes the trace large is not so much the fields, but the overhead required to store the schema with each record, and the usage of base64 encoding for lack of a bytes type and u128 type. Just eyeballing it, it looks like somewhere between 50-85% of the space is wasted to those.

I don't suggest developing our own binary format. I do suggested storing the data using Flatbuffers instead of Json, if using blob storage in Postgres.

> fit that into a Postgres schema.

This is compelling. I don't know yet what feature set Postgres affords but it's likely just as good or better than FlatBuffers, and may give us better options about where to filter. I'll do some poking around and see what's available here.

That3Percent: > we could use a more fine grained key for the cache, for example `(block, signature, contract)` since the call triggers in a subgraph will always specify a signature and usually specify an address.

My worry for using `trace_filter` instead of `trace_block` and having a fine grained cache is that we'll never actually get a cache hit, except when rolling back due to blocks being reverted. Even then, the cache would seem dubious.

This worry is based on the following ideas about how a Graph Node works, which may be wrong:

- When a Graph Node indexes a subgraph, it plays blocks forward in time and doesn't tend to revisit the same block for the same subgraph.
- It is only necessary to retrieve the traces once when indexing a block.
- It is possible but uncommon for multiple subgraphs to be based on the same Ethereum contract.

One possible approach to resolving the issue of increasing cache hits while not storing a block would be for the `EthereumTraceCache` to know the full set of contracts and addresses that are required by all of the subgraphs that are currently being indexed. This might be a code smell depending on how the code is structured and whether there would be any code required to keep these in sync. Here when any contract from a block is requested the cache would prefetch all the contracts and addresses needed from that block in the same request and store them separately in Postgres.

rfcs/0002-tracing-cache.md
That3Percent: Will fix.

rfcs/0002-tracing-cache.md
leoyvens: What you're missing is that subgraphs get re-synced with small changes all the time, in this situation a fine grained cache would still get plenty of hits.

That3Percent: > What you're missing is that subgraphs get re-synced with small changes all the time, in this situation a fine grained cache would still get plenty of hits.

There it is. I knew there must be something I was missing and this is key. What causes a re-sync? Code changes for a subgraph?

rfcs/0002-tracing-cache.md
That3Percent: Yes. That the main purpose of the cache is for subgraph re-sync changes *everything*. The in memory cache was meant for requests for the same block across multiple subgraphs, but storing the trace for a whole block doesn't make sense anymore. Other parts of the RFC that don't make sense with this additional context include the futures re-use, the usefulness of zero-copy / skipping / lazy parse in the binary format, and... pretty much the whole RFC.

What's been nice about this RFC process is that although it's painful to re-write an RFC in response to missing information that pales in comparison to re-writing the code.

rfcs/0002-tracing-cache.md
leoyvens: Yes, a subgraph is identified by the IPFS hash of its contents, if any source file changes, the subgraph must sync again when deployed. Developing a subgraph involves a lot of re-deploying and waiting to sync.

rfcs/0002-tracing-cache.md
Jannis: That's (part of) the idea — thinking is cheaper than writing code 😄

rfcs/0002-tracing-cache.md
That3Percent: > Developing a subgraph involves a lot of re-deploying and waiting to sync.

Would it make sense then to have caching disabled for a production Graph Node?

leoyvens: Not at all, most of our users do their development iterations on our production infra.

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+  <dt>Date of approval</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Approved by</dt>
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Summary
+
+This RFC proposes the creation of a local ethereum tracing cache to speed up indexing of subgraphs which use block and/or call handlers.
```
Jannis: Here's comes the nit-picky Jannis (you'll get used to me): Can we spell ethereum `Ethereum` with a capital E? Mainly because it's its own name, like Greenpeace or GitHub.

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Motivation
+
+When indexing a subgraph that uses block and/or call handlers, it is necessary to extract calls from the trace of each block that a Graph Node indexes. It is expensive to acquire and process traces from ethereum nodes in both money and time.
```
Jannis: Same here

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+
+When developing a subgraph it is common to make changes and deploy those changes to a production Graph Node for testing. Each time a change is deployed, the Graph Node must re-sync the subgraph using the same traces that were used for the previous sync of the subgraph. The cost of acquiring the traces each time a change is deployed impacts a subgraph developer's ability to iterate and test quickly.
```
Jannis, line 31: Great summary! 👍

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Urgency
+
+None
+
+## Terminology
+
+Ethereum cache: The new API proposed here.
```
Jannis: Let's make new terms italic, so `_Ethereum Cache_` or `*Ethereum Cache*`.

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+
+There is an existing `EthereumCallCache` for caching `eth_call` built into Graph Node today. This cache will be extended to support traces, and renamed to `EthereumCache`.
+
+## Compatibility
+
+This change is backwards compatible. Existing code can continue to use the parity tracing API. Because the cache is local, each indexing node may delete the cache should the format or implementation of caching change. In this case of invalidated cache the code will fall back to existing methods for retrieving a trace and repopulating the cache.
+
+## Drawbacks and Risks
+
+Subgraphs which are not being actively developed will incur the overhead for storing traces, but will not ever reap the benefits of ever reading them back from the cache.
```
Jannis, line 51: This is a great point. We could eventually remove data from the cache (actually, both caches — the `eth_call` and the `trace_filter` one) if the contracts the data is for are no longer used by any subgraphs. At the same time we don't want to do this too eagerly because someone might deploy a subgraph again that will use the cached data. Then again, all it does is affect performance, so cleanups will not do much harm.

Let's keep it in the back of our minds for now. Having this note here is good.

rfcs/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,61 @@
...
+
+If this drawback is significant, it may be necessary to extend `EthereumCache` to provide a custom score for cache invalidation other than the current date. For example, `trace_filter` calls could be invalidated based on the latest update time for a subgraph requiring the trace. It is expected that a subgraph which has been updated recently is more likely to be updated again soon then a subgraph which has not been recently updated.
```
Jannis, line 53: @lutter may have some thoughts here, since he's been doing a bunch of work on cleaning up stale/old db data recently. 

I like the idea of having a "last requested" timestamp that could act as a cleanup threshold. Would there be much overhead in updating cached traces in the db every time we read them?

Jannis (changes requested):
Tiny changes, this is looking great!

rfcs/0002-ethereum-tracing-cache.md
That3Percent: > I like the idea of having a "last requested" timestamp that could act as a cleanup threshold. Would there be much overhead in updating cached traces in the db every time we read them?

This is already happening: [store.rs #L13](https://github.com/graphprotocol/graph-node/blob/1d4695a11a4043a5a29f3aba4ac381dc1c3d4e43/store/postgres/src/store.rs#L1351)

As one might assume, this table was added to prune old data: [db_schema.rs #L74](https://github.com/graphprotocol/graph-node/blob/1d4695a11a4043a5a29f3aba4ac381dc1c3d4e43/store/postgres/src/db_schema.rs#L47)

But, I couldn't find anything that was actually pruning the data.

Jannis: Ah, great. In that case my comment is obsolete.

I don't think we're cleaning the `eth_call` cache up yet. But that's ok. As long as we have the possibility to start doing it whenever we feel it becomes necessary, we're good. 😄

Jannis (approved):

Jannis:
@That3Percent Approved! Make sure to fill in the date and approver in the meta data at the top before merging. Thanks 😄

PR #6, opened 2019-12-13
Subgraph Composition

ghost:
The engineering plan only includes schema merging right now, as I'm still working through query execution.


engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
+# PLAN-0001: Subgraph Schema Merging for Subgraph Composition
```
Jannis: I believe this is `PLAN-0002`. Let's shorten the title also: `Subgraph Schema Merging`.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+  <dt>Author</dt>
+  <dd>Jorge Olivero</dd>
+
+  <dt>Implements</dt>
+  <dd><a href="../rfcs/0001-subgraph-composition.md">RFC-0001 Subgraph Composition</a></dd>
+
+  <dt>Engineering Plan pull request</dt>
+  <dd><a href="https://github.com/graphprotocol/rfcs/pull/5">Engineering Plan PR</a></dd>
+
+  <dt>Obsoletes (if applicable)</dt>
```
Jannis: Let's keep the `<dd>` element below this but either write `<dd>None</dd>` or `<dd>-</dd>`. No `dd` element breaks the markup.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+  <dt>Date of approval</dt>
+  <dd>TBD</dd>
+
+  <dt>Approved by</dt>
+  <dd>TBD</dd>
+</dl>
+
+## Summary
+
+Subgraph composition allows a subgraph to import types from a foreign subgraph. Imports are designed to be loosely coupled throughout the deployment and indexing phases, and tightly coupled during query execution for the subgraph.
```
Jannis: `a foreign` -> `another`

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+To generate an API schema for the subgraph, the local schema needs to be merged with each of the foreign schemas that it imports. The merging process needs to take the following into account:
```
Jannis: `local schema` -> `subgraph schema`
`foreign schemas that it imports` -> `the imported subgraph schemas`.

Jannis: (Reason: The RFC switched from `foreign` to `imported` everywhere.)

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+1. An imported schema not being available on the graph-node
```
Jannis: graph-node -> Graph Node

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+2. An imported schema without a definition for the type which the local subgraph imports
```
Jannis: `without a definition for the type which the local subgraph imports` -> `missing a type imported by the subgraph schema`.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+3. A merged and cached api schema for a subgraph imported by name is updated
```
Jannis: We're not merging API schemas but original schemas (or intermediate versions of them), so let's just say `A schema imported by subgraph name changes` (`is updated` is also fine).

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+4. Ability to tell which subgraph/schema each type belongs to
+
+## Implementation
+
+There are two important parts to schema merging:
+
+1. Merging the schema
```
Jannis: schema -> schemas

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+2. Invalidating a cache entry when a subgraph's dependency, referenced by name, is updated and remerging that schema
```
Jannis: Nothing above speaks about caching yet, so perhaps we should talk about that first? What is cached and why and how?

Jannis: So I'd say something like
```
## Implementation

The schema merging implementation consists of two parts:

1. A cache of subgraph schemas
  - Each schema is associated with a subgraph name and/or deployment ID
  - Each schema is associated with all its dependencies (subgraph names and/or deployment IDs)
  - Each schema is stored in its merged form
2. The actual schema merging logic

I'm making assumptions about the cache here, mainly because they are not specified well enough in this plan yet.

File: engineering-plans/0002-subgraph-schema-merging.md [hunk 10]
```diff
@@ -0,0 +1,118 @@
...
+
+### Schema Merging
+
+Add a `merged_schema(&Schema, HashMap<SchemaReference, Arc<Schema>) -> Schema` function to the `graphql::schema` crate which will add each of the imported 
+types to the proviced document with a @subgraphId diretive denoting which subgraph the type came from. The `api_schema` function will add all the necessary types and fields for the imported types without any changes.
```
Jannis: As we established in the call about importing/implementing interfaces, just `@subgraphId` will not be enough. We also need to store the original type names, because imports may be aliased. So I suggest a new `@importInfo`, `@subgraphInfo` or `@mergeInfo` directive.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+type A @entity @subgraphId("...") {
+	id: ID!
+	foo: B!
+}
+
+type B @entity @subgraphId("...") {
+	id: ID!
+	bar: String
+}
+```
```
Jannis: Can we use 2 spaces for the indentation in these snippets? And add an aliased import to the example?

engineering-plans/0002-subgraph-schema-merging.md
Jannis: I'd also like some more details to be included on how the merging algorithm works.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+After the schema document is merged, the `api_schema` function will be called.
+
+### Cache Invalidation
+
+In the initial implementation, the cache will need to check if an entry needs to be updated when it is accessed. To determine whether an entry in the schema cache has been invalidated, each subgraph schema it imports by name needs to be queried in the store for its most current version. The schema cache will maintain a record of the versions used for merging in each entry, and if the version previously used to merge is no longer current that entry needs to be invalidated an remerged before responding to the cache access.
```
Jannis: Can we specify this more explicitly (I already hinted at this earlier when trying to come up with a good introduction for the implementation section)? What does the cache look like, what data structures does it use, what does the invalidation check algorithm look like (e.g. in bullet points)?

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Tests
+
+1. Schemas are merged correctly when all schemas and imported types are available
+
+2. Placeholder types are properly inserted into the merged schema when a schema is not available
+
+3. Placeholder types are properly inserted into the merged schema when the relevant schemas are available but the types are not
+
+4. The cache is invalidated when the store returns an updated version of a cache entry's dependency
```
Jannis: We're writing a proper document here, can we add full stops at the end of sentences? 😄

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+}
+```
+
+After the schema document is merged, the `api_schema` function will be called.
+
+### Cache Invalidation
+
+In the initial implementation, the cache will need to check if an entry needs to be updated when it is accessed. To determine whether an entry in the schema cache has been invalidated, each subgraph schema it imports by name needs to be queried in the store for its most current version. The schema cache will maintain a record of the versions used for merging in each entry, and if the version previously used to merge is no longer current that entry needs to be invalidated an remerged before responding to the cache access.
+
+A more performant invalidation solution would be to have the cache maintain a listener notifying it every time a subgraph's current version changes. Upon receiving the notification it would scan the schemas in the cache for those which need to be remerged.
```
Jannis: That's a good idea; and I agree this is future work. I wonder if we should add a "Future Work" section to engineering plans? What do you think?

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+2. Placeholder types are properly inserted into the merged schema when a schema is not available
+
+3. Placeholder types are properly inserted into the merged schema when the relevant schemas are available but the types are not
+
+4. The cache is invalidated when the store returns an updated version of a cache entry's dependency
+
+## Migration
+
+Subgraph composition is an additive feature which shouldn't require a special migration plan.
```
Jannis: Should't or doesn't? 😉

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Documentation
+
+Documenation on https://thegraph.com/docs needs to outline:
```
Jannis: Typo: `Documenation` -> `Documentation`

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+1. The reserved _Schema_ type and how to define imports on it
```
Jannis: `.` at the end 😉

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+2. The semantics of importing by subgraph id vs. subgraph name, i.e. what happens when a subgraph imported by name removes expected types from the schema
```
Jannis: `id` -> `ID`. Full stop at the end.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+3. How queries are processed when imported subgraphs schemas or types are not available on the graph-node processing the query
+
+## Implementation Plan
+
+- Impelment the `merged_schema` function (2d)
+- Write tests for the `merged_schema` function (1d)
+- Integrate `merged_schema` into `Store::cached_schema` and update the cache to include the relevant information for imported schemas and types (1d)
+- Add cache invalidation logic to `Store::cached_schema` (2d)
```
Jannis: Let's revisit this plan after the other points have been addressed. The plan is too vague for me still. Specifically the merging and caching algorithms are black boxes right now that I wouldn't be comfortable estimating yet.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Open Questions
+
+The execution of queries and subscriptions needs to be updated to leverage the types in a merged schema.
```
Jannis: We've been using bullet points for open points/questions so far, that would be good here too.

Jannis (changes requested):

engineering-plans/0002-subgraph-schema-merging.md
ghost: Yeah, I think it would be useful to make explicit. It's also a nudge to pare down your work to the essentials

Jannis (changes requested):
Quick comment: This is missing entries in `SUMMARY.md` and `approved.md` like these:

image

ghost:
@Jannis Comments addressed except for the one from this morning, addressing that now

ghost:
@Jannis I added the list item in the `SUMMARY.md`, it felt redundant though

ghost:
Ahh, the SUMMARY file is for mdBook

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,214 @@
...
+## Implementation
+
+The schema merging implementation consists of two parts:
+
+1. A cache of subgraph schemas
+2. The schema merging logic
+
+### Schema Merging
+
+Add a `merged_schema(&Schema, HashMap<SchemaReference, Arc<Schema>) -> Schema` function to the `graphql::schema` crate which will add each of the imported types to the provided document with a @subgraphId diretive denoting which subgraph the type came from.
```
Jannis: `diretive` -> `directive`

Jannis: `@subgraphId` in backticks.

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,214 @@
...
+
+The `HashMap<SchemaReference, Arc<Schema>` includes all of the schemas in the subgraph's import graph which are available on the Graph Node. For each `@import` directive on the subgraph, find the imported types by tracing their path along the import graph.
+
+- If any schema node along that path is missing or if the type is missing in the schema, add a type definition to the subgraphs merged schema with the proper name, a `@subgraphId("...")` directive (if available), and a `@placeholder` directive denoting that type was not found. 
+- If the type is found, copy it and add a `@subgraphId("...")` directive.
+- If the type is imported with the `{ name: "", as: "" }` format, the merged type will include an `@originalName("...")` directive preserving the type name from the original schema.
```
Jannis: Do we put (or have to put) the subgraph name somewhere if the import was from a subgraph name?

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,214 @@
...
+
+Schema after calling `merged_schema`:
+
+```graphql
+type A @entity {
+  id: ID!
+  foo: B!
+}
+
+type B @entity @subgraphId("X") {
```
Jannis: It's actually `@subgraphId(id: "X")`, because directives require named arguments. You can tell because it breaks the syntax highlighting from here on. 😉

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,214 @@
...
+  id: ID!
+  bar: String
+}
+```
+
+After the schema document is merged, the `api_schema` function will be called.
+
+### Cache Invalidation
+
+For each schema in the cache, keep a vector of subgraph pointers containing an element for each schema in the subgraph's import graph which was imported by name and the subgraph_id which was used during the schema merge. When a schema is accessed from the schema cache (and possibly only if this check hasn't happened in the last N seconds), check the current version for each of these schemas and run a diff against the versions used for the most recent schema merge. If there are any new versions, re merge the schema.
```
Jannis: `subgraph_id` -> `subgraph ID`

engineering-plans/0002-subgraph-schema-merging.md
```diff
@@ -0,0 +1,214 @@
...
+
+Currently the `schema_cache` in the `Store` is a `Mutex<LruCache<SubgraphDeploymentId, SchemaPair>>`. A `SchemaPair` consists of two fields: `input_schema` and `api_schema`. To support the refresh flow, `SchemaPair` would be extended to be a `SchemaEntry`, with the fields `input_schema`, `api_schema`, `schemas_imported_by_name` (`Vec<(SubgraphDeploymentName, SubgraphDeploymentId)>`), and a `last_refresh_check` timestamp.
+
```
Jannis: I don't think `schemas_imported_by_name` is sufficient. Don't we also need to re-check imports from subgraph IDs that were not available in previous merges and later became available?

Perhaps it should be `schema_imports` or `schemas_imported` ?

Jannis (changes requested):
Looking pretty good! Minor remarks, but you can go ahead and transfer the tasks over to Linear! 👍

lutter:
There's one issue here when it comes to querying data from the `Store`: with prefetching, we sometimes join two tables for a query `parents { children { id } }` when we get the `children`. With composition, it is possible that `parents` and `children` live in different subgraphs, something that `EntityQuery` can currently not express. I see two options for making this work:

1. Rework the prefetch code to avoid joins (we have the data we need from the joins already in memory when we query for `children`)
1. Change `EntityQuery` and the code that runs these queries to allow mixing entities from multiple subgraphs

Since, at some point, we will also want to be able to filter by child fields (`parents(where: {children.color: green}) { id }`) we should go with the second option, since for that we will need to join `parents` and `children` when we look up the `parents`.

engineering-plans/0002-subgraph-schema-merging.md
ghost: There's no reason I can think of right now, at least not until we want to do incremental remerges. I'll ponder this while implementing

Jannis:
@lutter Have your concerns been addressed yet? It reads more like it is relevant to query execution rather than merging, i.e. #12 rather than this PR.

PR #7, opened 2019-12-19
PLAN-0003: Engineering plan for removing JSONB storage

lutter:


engineering-plans/0003-remove-jsonb-storage.md
```diff
@@ -0,0 +1,125 @@
...
+
+  <dt>Date of submission</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Date of approval</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Approval needed</dt>
+  <dd>Yaniv/Jess/Eva on general approach and help with notifying users</dd>
+  <dd>Core devs on migration strategy</dd>
```
Jannis: Heh, this is not how `dl`/`dt`/`dd` work. You can only have one `dd` element after a `dt`. `dl` is like a glossary where you have pairs of definition topics (`dt`) and definition descriptions (`dd`).

engineering-plans/0003-remove-jsonb-storage.md
```diff
@@ -0,0 +1,125 @@
...
+## Implementation Plan
+
+_No estimates yet as we should first agree on this general course of
+action_
+
+* Notify hosted users to update their subgraph or have it deleted by date X
+* Mark JSONB storage as deprecated and announce when it will be removed
+* Provide tool to ship with `graph-node` to delete unused deployments and
+  unneeded subgraphs
+* Add affordance to not version entities to relational storage code
```
Jannis, line 111: If we make this per-entity, then we could also use it for the network indexer.

engineering-plans/0003-remove-jsonb-storage.md
```diff
@@ -0,0 +1,125 @@
...
+* Write SQL migrations to create new subgraph metadata schema and copy
+  existing data
```
Jannis, line 113: Do we have a rough idea of how long this migration would run for (thinking about downtime here).

Jannis (changes requested):
I'm on board with the plan. Just a couple questions and a comment about the HTML markup at the top.

engineering-plans/0003-remove-jsonb-storage.md
lutter: The W3C thinks otherwise: https://dev.w3.org/html5/html-author/#the-dl-element :smiley:

engineering-plans/0003-remove-jsonb-storage.md
lutter: Yes, it will have to be per entity, since we have metadata that is unversioned (belonging to a static data source) and versioned (belonging to a dynamic data source), e.g., `EthereumContractSource`

engineering-plans/0003-remove-jsonb-storage.md
lutter: It should be pretty quick; right now, we have around 500k metadata entities, and all we need is a fairly straightup copy. I think that it should be possible to do that in a couple minutes or so.

lutter:
@JessTaDa @yanivtal Could you have a look and comment that you are ok with the parts of this RFC that requires communicating with users (we'll have to tell them to redeploy or lose their subgraph by a certain date)? And what do you think a reasonable cadence is (send email on this date; expect them to respond by this date)

engineering-plans/0003-remove-jsonb-storage.md
Jannis: :mindblown:

Jannis: Having written HTML since 2000 or so... how did I not know this?!

Jannis (approved):

lutter:
Rebased on top of latest master

JessTaDa:
@lutter this sounds like a good plan, assuming that users are able to be contacted via the same emails we have in Grafana. I can also send a few reminder emails once we approach a decided date.

PR #9, opened 2019-12-20
Ethereum Tracing Cache: Added Engineering Plan

That3Percent:


engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+A temporary benchmark will be added for indexing a simple subgraph which uses call handlers. The benchmark will be run in these scenarios:
+* Sync before changes
+* Re-sync before changes 
+* Sync after changes
+* Re-sync after changes
+
+
+## Migration
+
+The old `eth_call_cache` and `eth_call_meta` tables will be deleted.
```
Jannis: We've already got 14,093,792 calls cached in these tables and a number of important subgraphs take e.g. 45 minutes instead of 8 hours thanks to this. I'd rather not throw this data away.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+Implements RFC-0002: Ethereum Tracing Cache
+
+## Implementation
+
+These changes happen within `ethereum_adapter.rs`, `store.rs` and `db_schema.rs`.
+
+### ServiceCache trait
+EthereumCallCache will be renamed to ServiceCache and have roughly this signature:
+```rust
+trait ServiceCache {
```
Jannis: I get the sentiment of wanting a more reusable cache, but I feel there is a benefit to the self-documenting schema of proper Postgres tables and a more specific API. What I do like here is the switch to futures and the callback to perform work on a cache miss.

We may not use this information yet, but the `block_number` and `contract_address` may become important debugging and cache cleanup tools. Opaque byte data lacks hints like that and, like with Ethereum's byte encoding, requires the caller to know what the structure of the data is, rather than having it manifested in a schema.

I'd prefer something like

```rust
struct EthCallKey {
    contract_address: ethabi::Address,
    block: EthereumBlockPointer,
    encoded_call: &[u8],
}

struct EthCallResult {
    contract_address: ethabi::Address,
    block: EthereumBlockPointer,
    encoded_call: &[u8],
    return_value: &[u8],
}
  
trait EthereumCache {
    fn get_eth_call(
        &self,
        key: EthCallKey,
        f: impl Future<Output = Result<EthCallResult>, Error>>,
    ) -> Box<dyn Future<Output = Result<EthCallResult, Error>>>;

    fn get_trace_filter(
        &self,
        key: TraceFilterKey,
        f: impl Future<Output = Result<TraceFilterResult>, Error>>,
    ) -> Box<dyn Future<Output = Result<TraceFilterResult, Error>>>;
}
```

Plus `try_store_eth_call` / `try_get_eth_call`, `try_store_trace_filter` / `try_get_trace_filter` methods.

That way we can keep the `eth_call_data` table and just add an `eth_trace_filter_data` table alongside it.

Jannis (changes requested):
@leoyvens What do you think about the more generalized cache API and schema? My concern is that it's going a little too far and that we're losing "concreteness" and important hints.

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: (Note sure the type names are the best but you get the idea.)

Jannis: At the same time, I see that the last accessed hint may be sufficient for cleaning up the cache periodically. As you can tell, I'm torn. 😄

That3Percent:
> @leoyvens What do you think about the more generalized cache API and schema? My concern is that it's going a little too far and that we're losing "concreteness" and important hints.

@Jannis @leoyvens The idea behind generalization here is that we're probably going to be wanting to use this cache for a lot more soon. In particular, multi-blockchain may have a number of new requests that would benefit from caching. The less special casing that needs to be added to the `ServiceCache` trait the better. A new method, for each call to be cached will get messy here.

There are a couple of quick wins that may be a good middle ground. Just keeping the cache id in the schema is enough to recover information about what the opaque bytes are, or to blow away an entire cache which is no longer needed.

Alternatively we could try to augment the trait to make it super configurable (eg: generic over the database table). I'm not super excited about that, I imagine the generics would be more messy than the value they provide.

Jannis:
Yeah, they would be. The more I think about it the more I agree with the proposal. I'd like to hear Leo's opinion as well though, as he wrote the initial call cache.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Depending on how the conversation about generality shakes out, we can migrate these if necessary or won't need to if the schema doesn't change.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+    fn try_store(key: [u8; 16], data: &[u8]) -> Box<dyn Future<Output=Result<(), Error>>>;
+    fn try_get(key: [u8; 16]) -> Box<dyn Future<Output=Result<Option<Vec<u8>, Error>>>>;
+}
+```
+
+This signature aims to fix 2 problems:
+  * Today, EthereumCallCache get_call is blocking, and used by code which returns a Future. This is [not recommended for async code](https://doc.rust-lang.org/stable/std/future/trait.Future.html#runtime-characteristics) and can cause problems for both runtimes and future combinators. The new signature allows for the database work to be migrated to use non-blocking apis, or to be spawned in a threadpool.
+  * This signature moves the `get or call then set` logic into the trait where it only need be written once, rather than everywhere it is used. This also allows for certain optimizations, like only calculating the hash of the arguments once.
+
+The Hash trait would be from [digest-hash](https://crates.io/crates/digest-hash/0.3.0) allowing tiny_keccak use for data structures, taking over the roll that `contract_call_id` fills today.
```
Jannis: `std::Hash` doesn't allow that? (https://doc.rust-lang.org/std/hash/index.html)

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: I wonder if we could migrate these somehow. Perhaps @lutter has some ideas.

The ideal way of doing it would be with a Diesel migration — we'll need one to replace the current tables with the new ones anyway — but I'm not sure we can encode the arguments in the same way we'd do it on the Rust side? We're also looking at migrating 14M entries which, depending on how fast a migration like that runs, could mean downtime for minutes.

Perhaps the best approach for this is to pull the data out with a script / simple Rust utility, and put it back into the new tables. We could pull down the data first, then roll out the new cache that drops the old tables, migrate the data offline and upload it to the new tables once that's done.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Summary
+
+Implements RFC-0002: Ethereum Tracing Cache
+
+## Implementation
+
+These changes happen within `ethereum_adapter.rs`, `store.rs` and `db_schema.rs`.
+
+### ServiceCache trait
```
Jannis: Should we just call this `IndexingCache` or something? I'm not sure what the `Service` stands for. 😉

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+
+    // Required functions
+    fn try_store(key: [u8; 16], data: &[u8]) -> Box<dyn Future<Output=Result<(), Error>>>;
+    fn try_get(key: [u8; 16]) -> Box<dyn Future<Output=Result<Option<Vec<u8>, Error>>>>;
+}
+```
+
+This signature aims to fix 2 problems:
+  * Today, EthereumCallCache get_call is blocking, and used by code which returns a Future. This is [not recommended for async code](https://doc.rust-lang.org/stable/std/future/trait.Future.html#runtime-characteristics) and can cause problems for both runtimes and future combinators. The new signature allows for the database work to be migrated to use non-blocking apis, or to be spawned in a threadpool.
+  * This signature moves the `get or call then set` logic into the trait where it only need be written once, rather than everywhere it is used. This also allows for certain optimizations, like only calculating the hash of the arguments once.
```
Jannis: 💯

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+- [ ] Run benchmarks
+- [ ] Migrate/delete old database schema
+
+
+**Estimates:**
+- (1) Create benchmarks
+- (2) Refactor existing `EthereumCallCache` - into `ServiceCache` trait
+- (1) Use `ServiceCache` trait for traces
+- (0.5) Run benchmarks
+- (0.5) Migrate/delete old database schema
```
Jannis: Can we add these estimates to the implementation plan items, like so?

```
- (1) Create benchmarks
- (2) Refactor existing `EthereumCallCache` into `ServiceCache` trait
...

Total: 5 days 
```

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,122 @@
...
+
+Total: 5
+
+**Phases:**
+None
+
+**Integration:**
+Potential integration after refactoring existing eth_cache to use new trait.
+
```
Jannis: Let's drop these two items since they are not needed.

Jannis: I think the work is compact enough to not require more than one (final) integration.

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: If we write an offline migration script, then I suspect this will make the migration a 2 day task (development plus actual migration time).

Jannis (changes requested):

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: @Jannis We could preface the migration with a one-time cleanup of records that haven't been accessed in a while. Depending on the distribution, that may be a significant cleanup.

I like the script idea. Because this is just a cache there is no reason the migration has to be complete when the process starts. We could even do it ahead of time.

leoyvens:
About blocking, imo the strategy we should take while we don't have an async DB interface is not to make store methods assume they're being run within a tokio runtime or make them return futures by running the queries on a separate threadpool, but to spawn our top-level tasks that use the DB (which is almost all of them) with `spawn_blocking`. I'll do this as part of the tokio 0.2 upgrade. So blocking within the store implementation should not be a concern.

The trait interface looks good, though I don't think `fn get<T: Deserialize>` is going to work, that `T` should be an associated type and I don't think `Deserialize` is a necessary bound.

About the current DB tables, `block_number` we really just threw it in because it's small and we could. But `eth_call_meta` having the contract address as a key really is an optimization, since it reduces the amount of writes needed to update `accessed_at` and avoids duplicating every cache key. I do not think it's worthwhile to drop that table for the sake of generalization. Putting all caches in a single DB table would be detrimental for performance since the table and indexes will be bloated, though @lutter will know more on that.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: > `std::Hash` doesn't allow that? (https://doc.rust-lang.org/std/hash/index.html)

No, and for good reason. Originally I thought a good way to approach this would be to write an adaptor allowing `std::Hash` to be used with `digest::Digest` generally. Then we could just do `#[derive(Hash)]` and life would be easy.

But.. the implementation of Hash only guarantees stability within a single process and is not for inter-process use. This applies not just to the `Hasher` trait but the `Hash` trait as well. One issue is that some data types like `usize` are not portable. Another issue is that some data types like `*const u8` _should_ provide an implementation for `std::Hash` but not be used with `digest::Digest` ever. Another issue is that the compiler is free to re-order the fields in `#[derive(Hash)` each time that it compiles. Another is that `#[derive(Hash)]` or some implementation is free to inject process specific random data, like the `ProcessId` on Windows, or just some randomly generated number on startup, or `std::any::TypeId` because none of those break the invariant that the hash will be stable for the given process.

If we found ourselves needing this a lot, we might contribute back a `#[derive(Hash)]` for `digest-hash`. But, I don't think we need it frequently enough for that to be a big win on time.

Jannis: Fun. Ok, makes complete sense now, thanks for the insights!

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: > Should we just call this `IndexingCache` or something? I'm not sure what the `Service` stands for. 😉

`Service` stands for service in the micro-service sense.  We are making some remote call that needs to be cached. I'm not sure what `Indexing` stands for. But, I'm amenable to whatever name.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: > Can we add these estimates to the implementation plan items, like so?
> 
> ```
> - (1) Create benchmarks
> - (2) Refactor existing `EthereumCallCache` into `ServiceCache` trait
> ...
> 
> Total: 5 days 
> ```

Yeah, I feel that there's no need for a separate Implementation Plan and Estimates. Maybe more complicated plans need that I'm not sure.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: > I think the work is compact enough to not require more than one (final) integration.

Agreed.

leoyvens:
I think the implementors of the trait should have access to the DB so there can be separate tables with different scheams for each cache. So the trait interface would also need a DB connection.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: > About blocking, imo the strategy we should take while we don't have an async DB interface is not to make store methods assume they're being run within a tokio runtime or make them return futures by running the queries on a separate threadpool, but to spawn our top-level tasks that use the DB (which is almost all of them) with `spawn_blocking`. I'll do this as part of the tokio 0.2 upgrade. So blocking within the store implementation should not be a concern.

I don't think this is enough. The issue is that future combinators like `select!` rely on individual futures being not blocking, not just the tasks for the runtime. For example, you may attempt to use `select!` to add a timeout to a future. If the future is blocking the timeout will never fire because `select!` calls `poll` in sequence and can never call `poll` for the second timeout if the first is blocked.

That3Percent (commented):
> About blocking, imo the strategy we should take while we don't have an async DB interface is not to make store methods assume they're being run within a tokio runtime or make them return futures by running the queries on a separate threadpool, but to spawn our top-level tasks that use the DB (which is almost all of them) with `spawn_blocking`. I'll do this as part of the tokio 0.2 upgrade. So blocking within the store implementation should not be a concern.

I don't think this is enough. The issue is that future combinators like select! rely on individual futures being not blocking, not just the tasks for the runtime. For example, you may attempt to use select! to add a timeout to a future. If the future is blocking the timeout will never fire because select! calls poll in sequence and can never call poll for the second timeout if the first is blocked.

leoyvens:
@That3Percent That's a good point. We can start having new store methods run queries on a separate thread pool and return futures, and consider changing the existing ones at some point.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+Implements RFC-0002: Ethereum Tracing Cache
+
+## Implementation
+
+These changes happen within or near `ethereum_adapter.rs`, `store.rs` and `db_schema.rs`.
+
+### Limitations
+The problem of reorg turns out to be a particularly tricky one for the cache, mostly due to ranges of blocks being requested rather than individual hashes. To sidestep this problem, only blocks that are older than the reorg threshold will be eligible for caching.
+
+This workaround has an intended side effect. The target beneficiary of the workflow involving re-sync of recently deployed subgraphs. It may be undesirable to cache subgraphs which were not recently deployed and may never be re-synced. Such a subgraph would only incur the cost of caching, but never reap the benefits of reading from the cache. By only caching blocks outside the reorg threshold, only the subgraphs which have been recently deployed and are the most likely to benefit from caching are in fact cached.
```
Jannis: I may be misunderstanding, but the reorg threshold has nothing to do with how recently a subgraph was deployed.

The property you're describing stems from the fact that the cache is only populated when indexing a subgraph. Only new versions index all the way from the genesis block to the latest head, meaning that only new subgraphs (deployed after the feature is rolled out) populate the cache. Old subgraphs that exist today and may never be touched again will remain untouched.

Regardless of when a subgraph is deployed though, we never know how likely a new version of that subgraph is deployed again in the future. There will be a lot of cases where we populate the cache for nothing (e.g. someone creates a subgraph for testing but never does anything with it after it has indexed). That's why I think it's important that we implement the ability to invalidate parts of this cache periodically.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+}
+```
+
+#### Block schema
+
+Each cached block will exist as its own row in the database. Here is the schema for that cache:
+
+```rust
+  id -> (Nullable<Bytea>, Integer), // (contract_address, block_number)
+  trace -> Jsonb // Or Bytea
```
Jannis: I'd split the "id" across two columns (`address` and `block_number`) and define a multi-column index for efficiently looking traces up by a pair of address (optional) and block number.

Seeing how https://tomusdrw.github.io/rust-web3/web3/types/struct.Trace.html implements `serde::Deserialize`, JSONB would give us a trivial way of writing/reading traces. I think the `trace` column is actually `traces`, because it's all (matching) traces for the block.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+    traces: Vec<Trace>,
+}
+
+pub trait EthereumTraceCache: Send + Sync + 'static {
+    /// Attempts to retrieve traces from the cache. Returns ranges which were retrieved.
+    /// The results may not cover the entire range of blocks. It is up to the caller to decide
+    /// what to do with ranges of blocks that are not cached.
+    fn get_cache(contract_address: Option<H160>, blocks: Range<u64>
+        ) -> Box<dyn Future<Output=Result<Vec<TracesInRange>, Error>>>;
+    fn append_cache(contract_address: Option<H160> traces: Vec<TracesInRange>);
```
Jannis: I'm not too happy with the method names here. `get_cache` suggests that a cache is returned (rather than the method being called against a cache) and `append_cache` suggests a cache is added to the end of a list or queue of caches. 😉 

How about something like `traces_for_blocks` (for the getter) and `cache_traces`/`add` (for the setter)?

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+```rust
+  id -> (Nullable<Bytea>, Integer), // (contract_address, block_number)
+  trace -> Jsonb // Or Bytea
+```
+
+Traces will be stored as Jsonb or Bytea using MessagePack, whichever is easier.
+
+#### Valid Cache Range
+Because of the absence of trace data for a block is a valid cache result, the database must maintain a data structure indicating which ranges of the cache are valid.
```
Jannis: `Because of` -> `Because`

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+This is the schema for that structure:
+```rust
+start_block -> Integer
+end_block -> Integer
+contract_address -> Nullable<Bytea>
+```
+
+When inserting data into the cache, removing data from the cache, or reading the cache, a serialized transaction must be used to preserve atomicity between the valid cache range structure and the cached blocks. Care must be taken to not rely on any data read outside of the serialized transaction, and for the extent of the serialized transaction to not span any async contexts that rely on any `Future` outside of the database itself. The definition of the `EthereumTraceCache` trait is designed to uphold these guarantees.
```
Jannis, line 88: Just to make sure I understand it correctly: we can evict data from the cache but that may mean that we have to adjust the start or end block, right? E.g. when the current valid range is `[0, 100)` and we decide to remove the first 50 blocks, we would have to update the range to `[50, 100)`.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+In order to preserve space in the database, whenever the valid cache range is added it will be added such that adjacent and overlapping ranges are merged into it.
+
+### Cache usage
+The primary user of the cache is `EtheriumAdapter<T>` in the `traces` function. 
+
+The correct algorithm for retrieving traces from the cache is surprisingly nuanced. The complication arises from the interaction between multiple subgraphs which may require a subset of overlapping contract addresses. The rate at which indexing proceeds of these subgraphs can cause different ranges of the cache to be valid for a contract address in a single query.
+
+We want to minimize the cost of external requests for trace data. It is likely that it is better too...
```
Jannis: too -> to

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+ * Make fewer requests
+ * Not ask for trace data that is already cached
+ * Ask for trace data for multiple contract addresses within the same block when possible.
```
Jannis, line 100: Note: we may be able to do this in the cache but Parity's `trace_filter` method only supports filtering by one target address at a time (https://wiki.parity.io/JSONRPC-trace-module#trace_filter).

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+ There is one flow of data which upholds these invariants. In doing so it makes a tradeoff of increasing latency for the execution of a specific subgraph, but increases throughput of the whole system.
+
+ Within this graph:
+ * Edges which are labelled refer to some subset of the output data.
+ * Edges which are not labelled refer to the entire set of the output data.
+ * Each node executes once for each contiguous range it. That is, it merges all incoming data before executing, and executes the minimum possible times.
```
Jannis: Something's up with this sentence: `Each node executes once for each contiguous range it.` 😉

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Benchmark
+A temporary benchmark will be added for indexing a simple subgraph which uses call handlers. The benchmark will be run in these scenarios:
+* Sync before changes
+* Re-sync before changes 
+* Sync after changes
+* Re-sync after changes
+
+### Ranges
+Due to the complexity of the resource minimizing data workflow, it will be useful to have mocks for the cache and database which record their calls, and check that expected calls are made for tricky data sets.
```
Jannis, line 158: I wonder if it would be worth implement a non-optimized version if the minimizing flow is complex. 

Either way, mockall's `#[automock]` should give us an easy way to mock things out.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Database
+A real database integration test will be added to test the add/remove from cache implementation to verify that it correctly merges blocks, handles concurrency issues, etc.
```
Jannis, line 161: Isn't clear to me yet what the concurrency risks are. Is the main problem to deal with that things may change between the time the valid cache range for an address (or no address) is determined and the actual traces are fetched from the cache?

Thinking about this makes me a little concerned about the amount of synchronization that'll be necessary between subgraph-indexing threads to ensure there are no race conditions. @lutter what do you think?

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+    /// Attempts to retrieve traces from the cache. Returns ranges which were retrieved.
+    /// The results may not cover the entire range of blocks. It is up to the caller to decide
+    /// what to do with ranges of blocks that are not cached.
+    fn get_cache(contract_address: Option<H160>, blocks: Range<u64>
+        ) -> Box<dyn Future<Output=Result<Vec<TracesInRange>, Error>>>;
+    fn append_cache(contract_address: Option<H160> traces: Vec<TracesInRange>);
+}
+```
+
+#### Block schema
```
Jannis, line 57: There is one more complicating factor that we ought to consider: we're supporting multiple Ethereum networks and will want to cache traces for all of them. That means block numbers will not be globally unique — the same block number on mainnet will refer to a different block than on one of the testnets.

The easiest way to to incorporate this is probably to add another column to the two proposed tables here: `network -> string` (e.g. `"mainnet"`, `"ropsten"`, `"rinkeby"`, `"goerli"`, `"poa-core"`). For efficient lookups, this column will have to be included in the multi-column index (see the next comment).

Jannis: This would require to either add a `network: String` (or similar) argument to the trace cache methods or to instantiate multiple `EthereumTraceCache`s, one for each network supported by the index node.

Alternatively to a `network` column in the tables, we could have completely separate tables, two for each network. @lutter, what would you prefer?

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Migration
+
+None
+
+## Documentation
+
+None, aside from code comments
+
+**Estimates:**
```
Jannis: Let's rename this to `## Implementation Plan`.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+  - (0.5) Trace Serialization/Deserialization
+  - (1.0) Ranges Cache
+  - (0.5) Concurrency/Transactions
+  - (0.5) Integration tests
+- Data Flow
+  - (3) Implementation
+  - (1) Unit tests
+- (0.5) Run Benchmarks
+
+Total: 8
```
Jannis: We may be underestimating the effort required for integration tests. Since we don't currently have tests that involve indexing subgraphs, this may not be trivial. But perhaps you already had an approach in mind for these?

Jannis (changes requested):

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+The problem of reorg turns out to be a particularly tricky one for the cache, mostly due to ranges of blocks being requested rather than individual hashes. To sidestep this problem, only blocks that are older than the reorg threshold will be eligible for caching.
+
+This workaround has an intended side effect. The target beneficiary of the workflow involving re-sync of recently deployed subgraphs. It may be undesirable to cache subgraphs which were not recently deployed and may never be re-synced. Such a subgraph would only incur the cost of caching, but never reap the benefits of reading from the cache. By only caching blocks outside the reorg threshold, only the subgraphs which have been recently deployed and are the most likely to benefit from caching are in fact cached.
+
+### EthereumTraceCache
+
+The implementation introduces the following trait, which is implemented primarily by `Store`.
+
+```rust
+use std::ops::Range;
```
leoyvens: It will be cool to start using proper range types for block ranges. This should be `InclusiveRange` since block ranges are always inclusive.

engineering-plans/0002-ethereum-tracing-cache.md
leoyvens: In SQL, you can't have a nullable component of a composite primary key afaik. So the block number can't be part of the primary key. It would be good to have the SQL migrations in the plan.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+When inserting data into the cache, removing data from the cache, or reading the cache, a serialized transaction must be used to preserve atomicity between the valid cache range structure and the cached blocks. Care must be taken to not rely on any data read outside of the serialized transaction, and for the extent of the serialized transaction to not span any async contexts that rely on any `Future` outside of the database itself. The definition of the `EthereumTraceCache` trait is designed to uphold these guarantees.
+
+In order to preserve space in the database, whenever the valid cache range is added it will be added such that adjacent and overlapping ranges are merged into it.
+
+### Cache usage
+The primary user of the cache is `EtheriumAdapter<T>` in the `traces` function. 
+
+The correct algorithm for retrieving traces from the cache is surprisingly nuanced. The complication arises from the interaction between multiple subgraphs which may require a subset of overlapping contract addresses. The rate at which indexing proceeds of these subgraphs can cause different ranges of the cache to be valid for a contract address in a single query.
```
leoyvens, line 95: The algorithm looks correct, but when you mention "multiple subgraphs" I think you mean multiple data sources within a subgraph. We wouldn't want to introduce synchronization across subgraphs for this.

leoyvens:
One thing that I find missing is consideration for the call signature as part of the cache key. If we only have the contract address, aside from the range, then as soon as we have one data source without a contract address we'll cache all traces ever.

Jannis:
@leoyvens I wonder if we should perhaps disallow data sources with call handlers but without address...

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I can add a cache invalidation section.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I'll use a multi-column index, and rename trace to traces.

That3Percent: Are there any necessary migrations when just adding new tables?

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Fixed

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Fixed

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: The plan does not have (as of yet) an eviction policy. I will add it. You are correct that an eviction will have to update the range. We are also updating the range when adding to the cache. In order to keep the database small, if [>=0, >100] is added and [>=100, >200] is added the record in the database will end up [>=0, >200].

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Fixed

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Each node executes once for each contiguous range of blocks. Fixed.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: The concern with concurrency is that two tables are being utilized which need to preserve atomicity (the table for the actual blocks, and the table for the cache). Because multiple subgraphs may include the same data sources and be indexed at the same time, they contend for the db.

I'm confident the use of serialized transactions as suggested by lutter will do the trick. But, it's worth making sure that the additional code paths introduced by serialized transactions (mainly retry failed transactions) are in fact tested because it may be some time before the code paths would be hit otherwise.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Good point, I'll add this.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: The only 'integration' tests I had in mind were just for using a real database rather than a mock to test the various transactions.

Having a full graph indexing test is possible, but would probably be slow. I see you did some work on this recently.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Good catch. I did not know that these were inclusive.

engineering-plans/0002-ethereum-tracing-cache.md
leoyvens: I just mean it would be nice to write out the `CREATE TABLE` commands.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I did mean multiple subgraphs. More than one subgraph may require the same data sources, and be indexed at the same time.

Eg:
Subgraph 1 requires contract address A for indexing.
Subgraph 1 has been deployed some time ago, and is up-to-date.
Subgraph 2 requires contract address A and contract address B.
Subgraph 2 is being developed.

When Subgraph 2 is deployed, it asks for A&B for some block range. It should get A from the cache, and B from Ethereum.

I don't think that ignoring concurrency is possible, so some kind of synchronization is necessary. The RFC includes the minimum amount of synchronization possible to avoid data corruption - namely the serialized transactions. The more complicated version of synchronization (eg: 1 & 2 ask for an uncached range of A at the same time merging to a single request) is not proposed by this RFC. What is here is more like optimistic concurrency.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I was thinking the same at first, that it we may be able to get away with some shortcuts to save on engineering time. The thing I'm trying to avoid at all costs is for the cost of the cache system to outweigh it's benefits. I see that happen a lot with performance work in the wild. This feature has been at risk for that since the start since we know that some data will never be read from cache.

There are some trivial and common cases I think where the cache can become a regression without at least most of this workflow. I'll know more when writing it for real.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: My understanding is that our `fn trace` has a parameter `addresses: Vec<H160>` which sometimes maps to `toAddresses` in this API, which takes an array.

About "sometimes maps to" - I'd like to change this to an `Option<Vec<_>>` to make it more clear, rather than having an empty `Vec` have a magic meaning.

That3Percent:
> One thing that I find missing is consideration for the call signature as part of the cache key. If we only have the contract address, aside from the range, then as soon as we have one data source without a contract address we'll cache all traces ever.

> @leoyvens I wonder if we should perhaps disallow data sources with call handlers but without address...

Right now, filtering by call signature is not something that's supported by the Parity tracing API. I can only assume what we're doing in this case right now is to get the trace for the whole block and then discarding some portion. The RFC handles this (poorly) by having the Nullable contract address mean the whole block is traced. If caching the whole block is bad, we could do more here to support a filtered block being cached.

I'm still of the position that having a call signature and no contract address is just asking for trouble, so I'm open to either the solution to just disable caching (dubious feature can run slow if it means happy path is fast), or making this not possible upstream.

engineering-plans/0002-ethereum-tracing-cache.md
leoyvens: I see, as long as the synchronization is at the DB level, and therefore it doesn't matter if the subgraphs are in the same node or separates nodes, this sounds good to me.

leoyvens:
> I wonder if we should perhaps disallow data sources with call handlers but without address...

I'd be ok with that. Or not caching these as @That3Percent suggested. I believe there are subgraphs that use this, but I don't think any production subgraph _should_ use this feature.

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: Ok, sounds good.

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: As per discussion earlier, it'll be fine to add a `network` column and include it in the multi-column index used in lookups. Postgres 11 will make it possible to partition the table according to this network column so that the lookup performance for a specific network will not be affected by the data for any other network.

@lutter Would it be too wasteful to use network names here? Should we use network IDs instead? (It would require keeping a map of network names to IDs around somewhere...).

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: Ah, ok, that makes it more realistic. It'd be fun to have a `graph test` integration test that indexes the same subgraph twice, once without a populated cache and then with a populated cache. We can work that out separately though. 😄

Jannis:
@leoyvens @That3Percent Simply bypassing the cache for data sources with call handlers (or block handlers w/ a `filter: call`) but no contract address sounds like the most simple and least painful approach. Let's do that for now.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I'm tempted to use Range in the API even if the Parity API uses InclusiveRange, if only because Range is so ubiquitous and well understood. It would be possible to convert at the boundary to adapt to the Parity API. My biggest concern is that would make migrating code a bit more error prone. Thoughts?

Jannis (changes requested):
Quick comment while go through the plan again:

Could you add a link to the plan to `SUMMARY.md` and `engineering-plans/approved.md`?

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+#### Block schema
+
+Each cached block will exist as its own row in the database in an `eth_traces_cache` table.
+
+```rust
+eth_traces_cache(id) {
+  id -> Integer,
+  network -> Text,
```
Jannis, line 64: For efficiency, we could limit the length of network names to e.g. 32 or 40 characters.

That3Percent:
> Quick comment while go through the plan again:
> 
> Could you add a link to the plan to `SUMMARY.md` and `engineering-plans/approved.md`?

Done

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,201 @@
...
+Additionally, there are some subgraphs which may require traces from all or a substantial number of blocks and don't make effective use of filtering. In particular, subgraphs which specify a call handler without a contract address fall into this category. In order to prevent the cache from bloating, any use of Ethereum traces which does not filter on a contract address will bypass the cache.
+
+### EthereumTraceCache
+
+The implementation introduces the following trait, which is implemented primarily by `Store`.
+
+```rust
+use std::ops::RangeToInclusive;
+struct TracesInRange {
+    range: RangeToInclusive<u64>,
```
Jannis: `RangeToInclusive` wouldn't allow to represent the `start_block` of the valid trace cache range.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,201 @@
...
+Because the absence of trace data for a block is a valid cache result, the database must maintain a data structure indicating which ranges of the cache are valid.
+
+This is the schema for that structure:
+```rust
+id -> Integer,
+network -> Text,
+start_block -> Integer,
+end_block -> Integer,
+contract_address -> Nullable<Bytea>,
+```
```
Jannis: I wonder if it wouldn't make sense to merge this into the `eth_traces_meta` table?

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,201 @@
...
+    G --> L
+    F --> M
+    G --> M
+ ```
+
+ 
+This construction is designed to make the fewest number of the most efficient calls possible. It is not as complicated as it looks. The actual construction can be expressed as sequential steps with a set of filters preceding each step.
+
+### Useful dependencies
+The feature deals a lot with ranges and sets. Operations like sum, subtract, merge, and find overlapping are used frequently. [nested_intervals](https://crates.io/crates/nested_intervals) is a crate which provides some of these operations.
```
Jannis, line 145: It might be worth pointing out that `nested_intervals` only works with `Range`, not `RangeInclusive`. But I'm sure we can implement any operations we need pretty easily, so this is not to worry about.

engineering-plans/0002-ethereum-tracing-cache.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Migration
+
+None
+
+## Documentation
+
+None, aside from code comments
+
+**Implementation Plan:**
```
Jannis: Let's make this `## Implementation Plan`, so it shows up as a proper section header in the document 😄

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: Yeah, that makes sense (regarding the `Option<Vec<_>>`). It's always good to be explicit.

Jannis (changes requested):
We're close! Just a few comments now. Since it's mostly about the details, do you want to go ahead already and copy the tasks and their estimates over to Linear?

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: @lutter Just said on slack that one should never limit the length of a VarChar. Perhaps he can elaborate.

One approach that works well in games is just to use a hash for any string that you want to use for compares/lookups. The collision probability for 20 networks with a 64 bit integer is 1 in 10^17 https://preshing.com/20110504/hash-collision-probabilities/ (There is a handy table at the end of this article)

I've applied this to hashmap structures and a few other places that required very fast text search based on eg: https://ourmachinery.com/post/minimalist-container-library-in-c-part-2/ In general, fixed size values and ints will be faster than strings.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Oops. `RangeInclusive`

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: It's pretty trivial to convert `RangeInclusive` to `Range` as compared to implementing these operations.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Really the most efficient thing, even more efficient than Postgres 11 hash table partitioning, is to just have different tables per network. I'm not sure what capabilities, if any, we have for creating tables on the fly though.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I don't understand.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: Done.

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: True :)

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: We do create tables dynamically for subgraphs (David can share more details). It's definitely not hard. For a network like the Ethereum mainnet, we could create tables prefixed with  `ethereum_mainnet_`. I'm not sure it's worth doing here (yet).

As for hashing to e.g. a 64 bit integer... even the smallest probability makes me uneasy because under no circumstances can we return traces for the wrong network. I can see it working wonders in combination with collision detection in a hash map/set though. It wouldn't be hard to maintain a table that maps network names to a unique number that we then use for the network column. 

Perhaps we can start without all this though, just use the network name and once we have 
everything implemented, we can refine it? What do you think?

engineering-plans/0002-ethereum-tracing-cache.md
Jannis: I was just thinking that
```rust
eth_traces_meta(id) {
  id -> (network, contract_address),
  network -> Text,
  contract_address -> Bytea,
  accessed_at -> Date,
  network -> Text,
  start_block -> Integer,
  end_block -> Integer,
}
```
could work as well. The start/end blocks are also meta data in a way. Does that make more sense?

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I'm all for getting this out the door.

engineering-plans/0002-ethereum-tracing-cache.md
That3Percent: I suppose that would work. I don't have any strong opinions on this part either way.

Jannis: Since there's already so many pieces to keep track of (not in this RFC but overall), I'd be in favor of using two instead of three tables, one for the traces themselves and one for the meta data and range information.

That3Percent: Ok. I'll make that adjustment.

Jannis (approved):
Looks good to me!

leoyvens (approved):

PR #10, opened 2019-12-24
Mutations

dOrgJelli:
A working prototype can be found here: https://github.com/dorgtech/the-graph-mutations-spec

### DIFF
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -607,9 +607,6 @@ The existing alternative that protocol developers are creating for dApp develope
 - **Should dApps be able to define event handlers for mutation events?**
   dApps may want to implement their own handlers for specific events emitted from mutations. These handlers would be different form the reducers, as we wouldn't want them to be able to modify the state. Instead they could store their own state elsewhere within the dApp based on the events.
 
-- **Should the resolvers module be ES5 compliant?**  
-  The prototype was originally developed under these conditions. ES5 compliance has since been abandoned as it has proven nearly impossible to successfully transpile all dependencies into a single monolithic module.
-
 - **Should the Graph Node's schema introspection endpoint respond with the "full" schema, including the mutations' schema?**
   Developers could fetch the "full" schema by looking up the subgraph's manifest, read the `mutations.schema.file` hash value, and fetching the full schema from IPFS. Should the graph-node support querying this full schema directly from the graph-node itself through the introspection endpoint?  
 
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -1,0 +1,71 @@
+# RFC-0003: Mutations
+
+<dl>
+  <dt>Author</dt>
+  <dd>dOrg: Jordan Ellis, Nestor Amesty</dd>
+
+  <dt>RFC pull request</dt>
+  <dd><a href="TODO">URL</a></dd>
+
+  <dt>Date of submission</dt>
+  <dd>2019-12-20</dd>
+
+  <dt>Date of approval</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Approved by</dt>
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Contents
+
+<!-- toc -->
+
+## Summary
+
+GraphQL Mutations allow you to add executable functions to your schema. Callers can invoke these functions using GraphQL queries. An introduction to how Mutations are defined and work can be found [here](https://graphql.org/learn/queries/#mutations). This RFC will assume the reader understands how to use GraphQL Mutations in a traditional Web2 application. Going forward we'll describe how Mutations can be added to The Graph's toolchain, and used to replace web3 write operations the same way The Graph has replaced Web3 read operations.
+
+## Goals & Motivation
+
+Currently, The Graph has created a read semantic layer that describes smart contract protocols, which has made it easier to build applications ontop of complex protocols. Since dApps have two primary interactions with web3 protocols (reading & writing), the next logical addition is write support.
+
+Protocol developers that currently use a subgraph still often publish a Javascript wrapper library for their dApp developers. This is done to help speed up dApp development and promote consistency with protocol usage patterns. With the addition of Mutations to the Graph Protocol's GraphQL tooling, Web3 reading & writing can now both be invoked through GraphQL queries. dApp developers can now simply refer to a single GraphQL schema that defines the entire protocol.
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
+
+## Terminology
+
+_Mutation_: GraphQL Mutation.
+_Resolver_: Resolver function that's mapped to a mutation.
+_Resolver State_: The resolver function's state (transactions sent, data logged, etc).
+_Optimistic Response_: A response given to the dApp that predicts what the outcome of the mutation's execution will be. If it is incorrect, it will be overwritten with the actual result.
+
+## Detailed Design
+
+This is the main section of the RFC. What does the proposal include? What are the proposed interfaces/APIs? How are different affected parties, such as users, developers or node operators affected by the change and how are they going to use it?
+
+TODO: design the developer flow, below it have a detailed spec of each aspect.
+
+## Compatibility
+
+Is this proposal backwards-compatible or is it a breaking change? If it is breaking, how could this be mitigated (think: migrations, announcing ahead of time like with hard forks, etc.)?
+
+TODO: no breaking changes will be introduced. This is an optional add-on to existing and new subgraphs.
+
+## Drawbacks and Risks
+
+Why might we _not_ want to do this? What cost would implementing this proposal incur? What risks would be introduced by going down this path?
+
+TODO: compat issues (ES5)
+
+## Alternatives
+
+What other designs have been considered, if any? For what reasons have they not been chosen? Are there workarounds that make this change less necessary?
+
+TODO: talk about existing alternatives, and how they're ineffecient.
+
+## Open Questions
+
+What are unresolved questions?
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -246,7 +246,7 @@ import {
   EventPayload,
   EventTypeMap,
   ProgressUpdateEvent
-} from "@graphprotocol/mutations-ts"
+} from "@graphprotocol/mutations"
 
 /// Mutation Resolvers
 const resolvers: MutationResolvers = {
@@ -380,7 +380,7 @@ NOTE: If the resolvers module exports a `stateBuilder`, it's expected that the m
 
 ### dApp Integration
 
-In addition to the resolvers module defined above, the dApp has access to a run-time API to help with the instantiation and execution of mutations. This package is called `@graphprotocol/mutations-ts` and is defined like so:
+In addition to the resolvers module defined above, the dApp has access to a run-time API to help with the instantiation and execution of mutations. This package is called `@graphprotocol/mutations` and is defined like so:
   - `createMutations` - Create a mutations interface which enables the user to `execute` a mutation query and `configure` the mutation module.  
     ```typescript
     interface CreateMutationsOptions<
@@ -427,7 +427,7 @@ For applications using Apollo and React, a run-time API is available which mimic
       OperationVariables
     } from "@apollo/react-common"
     import { MutationHookOptions } from "@apollo/react-hooks"
-    import { CoreState } from "@graphprotocol/mutations-ts"
+    import { CoreState } from "@graphprotocol/mutations"
 
     interface MutationResultWithState<TState, TData = any> extends MutationResult<TData> {
       state: TState
@@ -474,7 +474,7 @@ import {
   createMutations,
   createMutationsLink,
   MutationState
-} from "@graphprotocol/mutations-ts"
+} from "@graphprotocol/mutations"
 import {
   Mutation,
   useMutation
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -163,6 +163,19 @@ Mutation resolvers of kind `javascript/es5` take the form of an ES5 javascript m
     ```typescript
     import { GraphQLFieldResolver } from 'graphql'
 
+    interface MutationContext<
+      TConfig extends ConfigGenerators,
+      TState,
+      TEventMap extends EventTypeMap
+    > {
+      [prop: string]: any,
+      graph: {
+        config: ConfigProperties<TConfig>,
+        dataSources: DataSources,
+        state: StateUpdater<TState, TEventMap>
+      }
+    }
+
     interface MutationResolvers<
       TConfig extends ConfigGenerators,
       TState,
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -133,7 +133,7 @@ type Mutation {
 }
 ```
 
-`graph-cli` handles the combining, parsing, and validating of these two schemas. The `graph-cli` verifies that the mutations schema defines a `type Mutation`, that all of the mutations within it are defined in the resolvers module (see next section).  
+`graph-cli` handles the parsing and validating of these two schemas. It verifies that the mutations schema defines a `type Mutation` and that all of the mutations within it are defined in the resolvers module (see next section).  
 
 ### Mutation Resolvers
 
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -603,7 +603,20 @@ The existing alternative that protocol developers are creating for dApp develope
   Since mutations can contain many different steps internally, it would be ideal to be able to support continuing resolver execution in the event the dApp abruptly shuts down.
 
 - **How can dApps understand what steps a given mutation will take during the course of its execution?**
-  dApps may want to present to the user friendly progress updates, letting them know a given mutation is 3/4ths of the way through its execution (for example) and a high level description of each step. I view this as closely tied to the previous open question above, as we could support continuing resolver executions if we know what step it's currently undergoing. More research will be done, and a potential solution to solve both of these use cases will be presented at a later date.
+  dApps may want to present to the user friendly progress updates, letting them know a given mutation is 3/4ths of the way through its execution (for example) and a high level description of each step. I view this as closely tied to the previous open question above, as we could support continuing resolver executions if we know what step it's currently undergoing. A potential implementation could include adding a `steps: Step[]` property to the core state, where `Step` looks similar to:
+  ```typescript
+  interface Step {
+    id: string
+    title: string
+    description: string
+    status: 'pending' | 'processing' | 'error' | 'finished'
+    current: boolean
+    error?: Error
+    data: any
+  }
+  ```
+
+  This, plus a few core events & reducers, would be all we need to render UIs like the ones seen here: https://ant.design/components/steps/
 
 - **Should dApps be able to define event handlers for mutation events?**
   dApps may want to implement their own handlers for specific events emitted from mutations. These handlers would be different from the reducers, as we wouldn't want them to be able to modify the state. Instead they could store their own state elsewhere within the dApp based on the events.
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -113,7 +113,7 @@ resolvers:
   types: ./index.d.ts
 ```
 
-NOTE: `resolvers.types` is optional, and is only required if the resolvers module exports a state builder. More on this below.
+NOTE: `resolvers.types` is required. More on this below.
 
 ### Mutations Schema
 
@@ -383,13 +383,15 @@ export default {
 
 // Required Types
 export {
+  Config,
   State,
   EventMap,
   MyEvent
 }
 ```
 
-NOTE: If the resolvers module exports a `stateBuilder`, it's expected that the mutations manifest has a `resolvers.types` file defined. The following types are expected to be defined in the .d.ts type definition file:
+NOTE: It's expected that the mutations manifest has a `resolvers.types` file defined. The following types must be defined in the .d.ts type definition file:
+  - `Config`
   - `State`
   - `EventMap`
   - Any `EventPayload` interfaces defined within the `EventMap`
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -113,7 +113,7 @@ resolvers:
   types: ./index.d.ts
 ```
 
-NOTE: `resolvers.types` is optional, and is only required if the resolvers' module exports a state builder. More on this below.
+NOTE: `resolvers.types` is optional, and is only required if the resolvers module exports a state builder. More on this below.
 
 ### Mutations Schema
 
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -61,8 +61,8 @@ This is urgent from a developer experience point of view. With this addition, it
 * _Ext State Builder_: A state builder defined by the mutation developer. It's responsible for initializing the ext state properties, and processing the ext events with its reducers.  
 * _Mutations Config_: Collection of config properties required by the mutation resolvers. Also referred to in this document as "_Config_". All resolvers share the same config. It's passed to the resolver through the mutation context like so: `context.graph.config`.  
 * _Config Property_: A single property within the config (ex: ipfs, ethereum, etc).  
-* _Config Generator_: A function that takes a config value, and returns a config property. For example, "localhost:5001" as a config value gets turned into a new IPFS client by the config generator.
-* _Config Value_: An initialization value that's passed into the config generator. This config value is provided by the dApp developer.
+* _Config Generator_: A function that takes a config argument, and returns a config property. For example, "localhost:5001" as a config argument gets turned into a new IPFS client by the config generator.
+* _Config Argument_: An initialization argument that's passed into the config generator function. This config argument is provided by the dApp developer.
 * _Optimistic Response_: A response given to the dApp that predicts what the outcome of the mutation's execution will be. If it is incorrect, it will be overwritten with the actual result.  
 
 ## Detailed Design
@@ -102,6 +102,7 @@ dataSources: ...
 ```
 `mutations/mutations.yaml`
 ```yaml
+specVersion: ...
 repository: https://npmjs.com/package/...
 schema:
   file: ./schema.graphql
@@ -179,10 +180,10 @@ Mutation resolvers of kind `javascript/es5` take the form of an ES5 javascript m
     ```
   * `config: ConfigGenerators` - A collection of config generators. The config object is made up of properties, that can be nested, but all terminate in the form of a function with the prototype:
     ```typescript
-    type ConfigGenerator = (value: any) => any
+    type ConfigGenerator<TArg, TRet> = (arg: TArg) => TRet
 
     type ConfigGenerators = {
-      [prop: string]: (ConfigGenerator | ConfigGenerators)
+      [prop: string]: ConfigGenerator<any, any> | ConfigGenerators
     }
     ```
     See the example below for a demonstration of this.
@@ -302,17 +303,17 @@ const resolvers: MutationResolvers = {
 /// Config Generators
 const config: ConfigGenerators = {
   // These function arguments are passed in by the dApp
-  ethereum (provider: AsyncSendable): Web3Provider {
-    return new ethers.providers.Web3Provider(provider)
+  ethereum (arg: AsyncSendable): Web3Provider {
+    return new ethers.providers.Web3Provider(arg)
   },
-  ipfs (provider: string): IPFS {
-    return new IPFS(provider)
+  ipfs (arg: string): IPFS {
+    return new IPFS(arg)
   },
   // Example of a custom config property
   property: {
     // Generators can be nested
-    a: (value: string) => { },
-    b: (value: string) => { }
+    a: (arg: string) => { },
+    b: (arg: string) => { }
   }
 }
 
@@ -483,9 +484,9 @@ import { createHttpLink } from "apollo-link-http"
 
 const mutations = createMutations({
   mutations: myMutations,
-  // Config values, which will be passed to the generators
+  // Config args, which will be passed to the generators
   config: {
-    // Config values can take the form of functions to allow
+    // Config args can take the form of functions to allow
     // for dynamic fetching behavior
     ethereum: async (): AsyncSendable => {
       const { ethereum } = (window as any)
@@ -616,4 +617,4 @@ The existing alternative that protocol developers are creating for dApp develope
   I have not thought of a trustless solution to this, am curious if anyone has any ideas of how we could make this possible.  
 
 - **Will The Graph Explorer support mutations?**
-  We could have the explorer client-side application dynamically fetch and include mutation resolver modules. Configuring the resolvers module dynamically is problematic though. Maybe there are a few known config properties that the explorer client supports, and for all others it allows the user to input config values (if they're base types).  
+  We could have the explorer client-side application dynamically fetch and include mutation resolver modules. Configuring the resolvers module dynamically is problematic though. Maybe there are a few known config properties that the explorer client supports, and for all others it allows the user to input config arguments (if they're base types).  
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -583,7 +583,7 @@ No breaking changes will be introduced, as mutations are an optional add-on to a
 
 ## Drawbacks and Risks
 
-I have some thoughts but they are rather verbose and tangential. Would love some feedback on this from others first.
+Nothing apparent at the moment.
 
 ## Alternatives
 
@@ -591,10 +591,11 @@ The existing alternative that protocol developers are creating for dApp develope
 
 ## Open Questions
 
+- **How can mutations pickup where they left off in the event of an abrupt application shutdown?**
+  Since mutations can contain many different steps internally, it would be ideal to be able to support continuing resolver execution in the event the dApp abruptly shuts down.
+
+- **How can dApps understand what steps a given mutation will take during the course of its execution?**
+  dApps may way to present to the user friendly progress updates, letting them know a given action is 3/4 of the way through execution (for example) and a high level description of each step. I view this as closely tied to the previous open question above, as we could support continuing resolver executions if we know what step it's currently undergoing. More research will be done, and a potential solution to solve both of these use cases will be presented at a later date.
+
 - **Should the resolvers module be ES5 compliant?**  
   The prototype was originally developed under these conditions. ES5 compliance has since been abandoned as it has proven nearly impossible to successfully transpile all dependencies into a single monolithic module.
-
-- **What paradigm should the mutation state follow?**  
-  One option is to have the resolver's call into a single interface that modifies the backing data. Whenever this data is modified, the entirety of it is passed to the dApp. The downside here is that the dApp doesn't know what has changed within the data, and is forced to represent it in its entirety in order to not miss anything.  
-
-  Another option is to implement something similar to Redux, where the resolvers fire off events with corresponding payloads of data. These events map to reducers, which take in this payload of data and decide what to do with it. The dApp could implement these reducers, and choose how it would want to react to the various events.
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -287,11 +287,6 @@ const resolvers: MutationResolvers = {
         }`
       })
 
-      // Fetch datasource addresses & abis from the subgraph manifest
-      const { MyContract } = context.graph.datasources
-      await MyContract.abi
-      await MyContract.address
-
       ...
     },
     async setEntityName (source: any, args: any, context: any) {
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -251,7 +251,9 @@ const CREATE_ENTITY = gql`
   }
 `
 
-// state === mutation state
+// exec: execution function for the mutation query
+// loading: https://www.apollographql.com/docs/react/data/mutations/#tracking-mutation-status
+// state: mutation state instance
 const [exec, { loading, state }] = useMutation(
   CREATE_ENTITY,
   {
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -137,7 +137,7 @@ type Mutation {
 
 ### Mutation Resolvers
 
-Each mutation within the schema must have a corresponding resolver function defined. Resolvers will be invoked by whatever engine executes the query. They are executed locally within the client application.
+Each mutation within the schema must have a corresponding resolver function defined. Resolvers will be invoked by whatever engine executes the mutation queries. They are executed locally within the client application.
 
 Mutation resolvers of kind `javascript` take the form of a javascript module. This module is expected to have a default export that contains the following properties:
   * resolvers - The mutation resolver functions.
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -485,7 +485,6 @@ For applications using Apollo and React, a run-time API is available which mimic
       TVariables
     > extends BaseMutationOptions<TData, TVariables> {
       mutation: DocumentNode
-      options: MutationHookOptions
       children: (
         mutateFunction: MutationFunction<TData, TVariables>,
         result: MutationResultWithState<TState, TData>
@@ -613,10 +612,13 @@ const [exec, { loading, state }] = useMutation(
 ```
 ```html
 // Use the Mutation JSX Component
-<Mutation mutation={CREATE_ENTITY} variables={{options: { name: "...", value: 5 }}}>
-{(exec, { loading, state }) => (
-  <button onClick={exec} />
-)}
+<Mutation
+  mutation={CREATE_ENTITY}
+  variables={{options: { name: "...", value: 5 }}}
+>
+  {(exec, { loading, state }) => (
+    <button onClick={exec} />
+  )}
 </Mutation>
 ```
 
--- a/rfcs/0003-mutations.md
+++ b/rfcs/0003-mutations.md
@@ -37,26 +37,39 @@ This is urgent from a developer experience point of view. With this addition, it
 
 ## Terminology
 
+* _Mutations_: Collection of mutations.  
 * _Mutation_: A GraphQL mutation.  
+* _Mutations Schema_: A GraphQL schema that defines a `type Mutation` that contains all mutations. Additionally, this schema can define other types to be used by the mutations, such as `input` and `interface` types.  
+* _Mutations Manifest_: A YAML manifest file that is used to add mutations to an existing subgraph manifest.  
+* _Mutation Resolvers_: Code module that contains all resolvers.  
 * _Resolver_: Function that is used to execute a mutation.  
-* _Mutation State_: The state of a mutation being executed.  
+* _Mutation State_: The state of a mutation being executed. It's passed to the resolver through the mutation context.  
+* _Mutation Context_: A context object that's created for every mutation that's executed. It's passed as an argument to the resolver.  
+* _Config_: Collection of config properties required by the mutation resolvers.  
+* _Config Property_: A single property within the config (ex: ipfs, ethereum, etc).  
+* _Config Generator_: A function that takes a config value, and returns a config property. For example, "localhost:5001" as a config value gets turned into a new IPFS client by the config generator.
+* _Config Value_: An initialization value that's passed into the config generator. This config value is provided by the dApp developer.
 * _Optimistic Response_: A response given to the dApp that predicts what the outcome of the mutation's execution will be. If it is incorrect, it will be overwritten with the actual result.  
 
 ## Detailed Design
 
-### Mutation Manifest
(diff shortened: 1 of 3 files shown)


rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+Protocol developers that currently use a subgraph still often publish a Javascript wrapper library for their dApp developers. This is done to help speed up dApp development and promote consistency with protocol usage patterns. With the addition of Mutations to the Graph Protocol's GraphQL tooling, Web3 reading & writing can now both be invoked through GraphQL queries. dApp developers can now simply refer to a single GraphQL schema that defines the entire protocol.
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
+
+## Terminology
```
namesty, line 38: Maybe here explain context object and mutationState object

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Drawbacks and Risks
+
+I have some thoughts but they are rather verbose and tangential. Would love some feedback on this from others first.
+
+## Alternatives
+
+The existing alternative that protocol developers are creating for dApp developers has been described above.
+
+## Open Questions
```
namesty, line 640: I think also an important thing to discuss here is that we currently depend on Apollo. 

Also what we talked about adding support for different frameworks, by releasing several framework specific libraries.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
+# RFC-0003: Mutations
+
+<dl>
+  <dt>Author</dt>
+  <dd>dOrg: Jordan Ellis, Nestor Amesty</dd>
+
+  <dt>RFC pull request</dt>
+  <dd><a href="TODO">URL</a></dd>
```
Jannis: This can be set to https://github.com/graphprotocol/rfcs/pull/10/ now (both `TODO` and `URL`).

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Contents
+
+<!-- toc -->
+
+## Summary
+
+GraphQL Mutations allow you to add executable functions to your schema. Callers can invoke these functions using GraphQL queries. An introduction to how Mutations are defined and work can be found [here](https://graphql.org/learn/queries/#mutations). This RFC will assume the reader understands how to use GraphQL Mutations in a traditional Web2 application. Going forward we'll describe how Mutations can be added to The Graph's toolchain, and used to replace web3 write operations the same way The Graph has replaced Web3 read operations.
```
Jannis: * Let's talk about `developers` rather than `you`. It's always unclear who "you" or "we" are in documents like this.
* Let's spell `mutations` without a capital `M`. It's just a regular word after all.
* Instead of `Going forward, we'll describe`, let's say `This proposal describes`
* Also, phrase things as if they were already accepted, so that when someone later reads this on https://graphprotocol.github.io/rfcs/, it doesn't read as if it's about the future.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Goals & Motivation
+
+Currently, The Graph has created a read semantic layer that describes smart contract protocols, which has made it easier to build applications ontop of complex protocols. Since dApps have two primary interactions with web3 protocols (reading & writing), the next logical addition is write support.
```
Jannis: Drop the `Currently, `

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+Protocol developers that currently use a subgraph still often publish a Javascript wrapper library for their dApp developers. This is done to help speed up dApp development and promote consistency with protocol usage patterns. With the addition of Mutations to the Graph Protocol's GraphQL tooling, Web3 reading & writing can now both be invoked through GraphQL queries. dApp developers can now simply refer to a single GraphQL schema that defines the entire protocol.
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
```
Jannis: `I`? 😉Please use passive language here (see my earlier comments).

Jannis: graphql -> GraphQL

Jannis: manaually -> manually

Jannis: user-friendly -> developer-friendly. Otherwise the user/developer perspectives get mixed up a bit too much in this paragraph.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Goals & Motivation
+
+Currently, The Graph has created a read semantic layer that describes smart contract protocols, which has made it easier to build applications ontop of complex protocols. Since dApps have two primary interactions with web3 protocols (reading & writing), the next logical addition is write support.
+
+Protocol developers that currently use a subgraph still often publish a Javascript wrapper library for their dApp developers. This is done to help speed up dApp development and promote consistency with protocol usage patterns. With the addition of Mutations to the Graph Protocol's GraphQL tooling, Web3 reading & writing can now both be invoked through GraphQL queries. dApp developers can now simply refer to a single GraphQL schema that defines the entire protocol.
```
Jannis: Can you include one or two examples of such wrapper libraries to strengthen the argument?

rfcs/0003-mutations.md
Jannis: What code generators are you thinking about? Would be nice to mention one or two (`such as ...`).

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
+
+## Terminology
+
+* _Mutation_: GraphQL Mutation.
```
Jannis: `GraphQL Mutation` -> `A GraphQL mutation`

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+Protocol developers that currently use a subgraph still often publish a Javascript wrapper library for their dApp developers. This is done to help speed up dApp development and promote consistency with protocol usage patterns. With the addition of Mutations to the Graph Protocol's GraphQL tooling, Web3 reading & writing can now both be invoked through GraphQL queries. dApp developers can now simply refer to a single GraphQL schema that defines the entire protocol.
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
+
+## Terminology
+
+* _Mutation_: GraphQL Mutation.  
+* _Resolver_: Resolver function that's mapped to a mutation.
```
Jannis: I'd call it `Function that is used to execute a mutation`.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Urgency
+
+From a developer experience point of view I see this as urgent because it eliminates the need for protocol developers to manaually wrap their graphql query interfaces alongside user-friendly write functions. Additionally from a user experience point of view, mutations provide a solution for optimistic UI updates, which is something dApp developers have been wanting for a long time. Lastly with the whole protocol now defined in GraphQL, existing application layer code generators can now be used to hasten dApp development.
+
+## Terminology
+
+* _Mutation_: GraphQL Mutation.  
+* _Resolver_: Resolver function that's mapped to a mutation.  
+* _Resolver State_: The resolver function's state (transactions sent, data logged, etc).
```
Jannis: `The resolver function's state` -> `The state of a mutation being executed`? Not sure. Something like that. What I'm getting at is that the state is related to a specific mutation being executed, not related to a function definition.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Terminology
+
+* _Mutation_: GraphQL Mutation.  
+* _Resolver_: Resolver function that's mapped to a mutation.  
+* _Resolver State_: The resolver function's state (transactions sent, data logged, etc).  
+* _Optimistic Response_: A response given to the dApp that predicts what the outcome of the mutation's execution will be. If it is incorrect, it will be overwritten with the actual result.  
+
+## Detailed Design
+
```
Jannis, line 69: A short paragraph that introduces the elements covered in the subsections would be good and that describes how they relate to the existing subgraph elements and how they fit into the subgraph development flow.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+### Mutation Manifest
+
+`subgraph.yaml`
+```yaml
+specVersion: ...
+...
+mutations:
+  specVersion: 0.0.1
```
Jannis: In the subgraph manifest we currently use `specVersion` only for the top level, and `apiVersion` at the mapping level (this would be resolvers here). Any reasons not to follow that same pattern?

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  repository: https://npmjs.com/package/...
+  schema:
+    file: ./mutations/schema.graphql
+  resolvers:
+    kind: javascript
```
Jannis: I'd say `javascript/es5` (is `es5` correct?) here, in case we later want to add support for e.g. `javascript/es6` or similar.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+...
+```
+`mutations/mutations.yaml`
+```yaml
+specVersion: 0.0.1
+repository: https://npmjs.com/package/...
+schema:
+  file: ./schema.graphql
+resolvers:
+  kind: javascript
```
Jannis: Same question about the `specVersion`/`apiVersion` here. I also notice that the one for the resolvers is missing.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  file: ./index.js
+```
+
+### Mutation Schema
+`schema.graphql`
```
Jannis, line 121: Could you add a short paragraph that describes what the mutations schema includes (or can include) and how it relates to the regular subgraph schema? Both appear here without any distinction or introduction.

Jannis: It would also be good to have a more elaborate description of how mutations are defined, what arguments and return values they can have and how these definitions relate to the resolvers.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+    options: MyEntityOptions!
+  ): MyEntity!
+
+  setEnityName(
+    entity: MyEntity!
+    name: String!
+  ): MyEntity!
+}
+```
+**NOTE:** GraphQL types from the subgraph's schema.graphql are automatically included in this file.
```
Jannis: This should be more than a note because it's an important part of the design. Let's make this part of the paragraphs above the schema.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Mutation Resolvers
+`mutations/index.js`
```
Jannis, line 246: In introductory/overview paragraph would be good here as well.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+```javascript
+const resolvers = {
+  Mutation: {
+    async createEntity (_, args, context) {
+      // Extract mutation arguments
+      const { name, value } = args.options
+
+      // Use configuration properties created by the
+      // config generator functions below
+      const { ethereum, ipfs } = context.graph.config
```
Jannis: We're using inconsistent terms for what the configuration is. Is it "config generator functions", is it "configuration setters"?

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+      // Fetch datasource addresses & abis
+      const { MyContract } = context.graph.datasources
+      await MyContract.abi
+      await MyContract.address
+
+      // Modify a state object, which relays updates back
+      // to the subscribed dApp
+      const { mutationState } = context.graph
+      mutationState.addTransaction("tx_hash")
```
Jannis: For ergonomic reasons, I'll through the idea into the ring of calling this just `state`.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  createMutations,
+  createMutationsLink,
+  useMutation
+} = require("@graphprotocol/mutations-ts")
+const myMutations = require("mutations-js-module")
+
+const mutations = createMutations({
+  mutations: myMutations,
+  subgraph: "my-subgraph",
+  node: "http://localhost:8080",
```
Jannis: What's this `node` used for? I assume it's the Graph node from which the subgraph's data sources are queried?

With the query engine, this will not be needed. The query engine needs some work though and we have to think a little about the integration of the two anyway. I'm ok keeping this argument for now but we should drop it rather soon.

rfcs/0003-mutations.md
Jannis: I think it's important that we make `createMutations` return something that's Apollo-agnostic. `createMutationLink` is just a relatively thin wrapper around it to integrate with Apollo, and thereby React.

I agree also that it would be good to mention how integration into at least one other framework (Vue?) could look like.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  mutation createEntity($options: MyEntityOptions) {
+    createEntity(options: $options) {
+      id
+      name
+      value
+    }
+  }
+`
+
+// state === resolver's state
```
Jannis: `resolver's state` -> `mutation state`?

Jannis: But I'd rather see some more human-readable comments explaining the different parts of the code snippet. 😉

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+  {
+    optimisticResponse: {
+      myEntity: {
+        id: "...",
+        name: "...",
+        value: 5,
+      }
+    },
+    update(proxy, { data }) {
+      // result = data.myEntity
```
Jannis: I'm not sure what the code in this comment is supposed to do.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+)
+
+// We can also utilize an optimistic response
+const [exec, { loading, state }] = useMutation(
+  CREATE_ENTITY,
+  {
+    optimisticResponse: {
+      myEntity: {
```
Jannis: I think this has to match the shape of the mutation query. See https://www.apollographql.com/docs/react/performance/optimistic-ui/.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+)
+```
+
+## Compatibility
+
+No breaking changes will be introduced, as mutations are an optional add-on to a subgraph.
+
+## Drawbacks and Risks
+
+I have some thoughts but they are rather verbose and tangential. Would love some feedback on this from others first.
```
Jannis: Why not put them into the open questions section?

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Alternatives
+
+The existing alternative that protocol developers are creating for dApp developers has been described above.
+
+## Open Questions
+
+- **Should the resolvers module be ES5 compliant?**  
+  We've been operating under this assumption while developing the prototype. We have since scrapped this requirement as it has proven nearly impossible to successfully transpile our own source, along with all our dependencies, into a single monolithic module. If anyone has experience doing this I would love chat!
```
Jannis: You'd think that it's pretty common and can be achieved with a transpiler and bundler, but I don't have much experience with that either. Bundling everything together into a single file at least should be no problem. Maybe ES2015 / ES6 is the better target nowadays?

Another question is the module system to use for exporting from the bundle (CommonJS, UMD, ...)?

@nenadjaja may have more experience with transpiling to older JS versions and bundling.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+- **How should the dApp configure the resolvers module?**  
+  The dApp knows best how to: connect to the various web3 networks, handle key signature requests, and all other user / dApp specific things. We need a way for the dApp to configure the resolvers in a specific way given the resolver's requirements (IPFS provider, Web3 provider, etc).
```
Jannis: Isn't that already covered in the design via the `config` API?

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,280 @@
...
+- **What paradigm should the resolver state follow?**  
+  One option is to have the resolver's call into a single interface that modifys the backing data. Whenever this data is modified, the entirety of it is passed to the dApp. The downside here is that the dApp doesn't know what has changed within the data, and is forced to represent it in its entirety in order to not miss anything.  
+
+  Another option is to implement something similar to Redux, where the resolvers fire off events with corresponding payloads of data. These events map to reducers, which take in this payload of data and decide what to do with it. The dApp could implement these reducers, and choose how it would want to react to the various events.
```
Jannis: There are a couple things I'm missing in the RFC so far:

1. What exactly does `createMutations` return? General, there's little in terms of a specification included aside from some code snippets.

2. What exactly does the mutation state API look like? When it comes to the paradigm, I don't think we're limited to a single one. To me, the ideal API would include both "static" state information as well as progress information like a progress percentage or a sequence of events. The big question and one that I was hoping we'd already be closer to is: what does this API look like? Here's a strawman:
    ```js
    type MutationEvent = TransactionEvent | ...
    type TransactionEvent = TransactionStartedEvent | TransactionCanceledEvent | TransactionConfirmedEvent | ...

    interface TransactionStartedEvent {
      transaction: Transaction
      ...
    }

    interface Transaction {
      id: string
      title: string
      ...
    }

    interface MutationState {
      progress: Number, // Between 0.0 and 1.0
      events: MutationEvent[],

      startTransaction({ id: string, title: string, ... })
      cancelTransaction(id: string, ...)
      confirmTransaction(id: string, ...)
      ...
    }
    ```
  what I'm not sure personally is what a good abstraction for transactions is across different chains is. I'm not overly familiar with meta transactions etc. so that makes it hard for me to come up with something on the spot. But this is what we need to think about — unless we l basically make `Transaction` an opaque structure with just an `id`, a human-readable `title`, perhaps a `link` to visit for details and an arbitrary `payload` or something.

Jannis (changes requested):

rfcs/0003-mutations.md
Jannis: https://www.tutorialspoint.com/babeljs/babeljs_transpile_es6_modules_to_es5.htm looks promising (webpack + babel with a single output file). I bet there are better guides out there though.

rfcs/0003-mutations.md
Jannis: Note: In the above strawman, the `progress` and `events` should be exposed to the user of `useMutation` but the `startTransaction` etc. functions should not. That could be achieved e.g. with
```ts
// Returned by `useMutation`
interface MutationState {
  progress: Number,
  events: MutationEvent[],
}

// Managed by the resolver
interface ManagedMutationState {
  state: MutationState

  startTransaction({ id: string, title: string, ... })
  cancelTransaction(id: string, ...)
  confirmTransaction(id: string, ...)
}
```

rfcs/0003-mutations.md
dOrgJelli: This endpoint is used to query the metadata subgraph. I don't think we can get rid of this here because we need to know the query endpoint's root URL so we can append `/subgraphs` to it.

In the case of using `createQueryEngine`, the developer could just pass in the mutations & config there like so:
```typescript
let engine = createQueryEngine({
  subgraph: 'subgraph/a',
  mutations: {
    module: myMutations,
    config: { ... }
  }

  subgraphNameResolver: ...,
  indexersResolver: ...,
  indexNodeResolver: ...,
  queryPlanner: ...,
  queryExecutor: ...,
})
```

Thoughts?

rfcs/0003-mutations.md
dOrgJelli: Ah okay that makes sense. If the structure of the mutation's object changes, the top level `specVersion` would be bumped, correct?

rfcs/0003-mutations.md
dOrgJelli: Yes ES5 was the target we originally talked about, but we encountered a lot of problems trying to transpile all of the prototype's node_modules to ES5. I'm thinking ES6 might be a more realistic target.

rfcs/0003-mutations.md
dOrgJelli: How does this look? Super simple, but I figured the individual parts are intro'd on their own.

rfcs/0003-mutations.md
dOrgJelli: How does it look now?

rfcs/0003-mutations.md
Jannis: Yep, I think that works.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+    kind: javascript/es5
+    file: ./mutations/index.js
+dataSources: ...
+...
+```
+
+Alternatively, the mutation manifest can be external like so:  
+`subgraph.yaml`
+```yaml
+specVersion: ...
```
Jannis, line 96: It may be worth noting that the `specVersion` here must match the `specVersion` of the main subgraph manifest. That way we tie the two together, which will make our live easier whenever we need to bump this version.

Jannis: Our validation in graph-cli should catch mismatches.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+  file: ./schema.graphql
+resolvers:
+  apiVersion: 0.0.1
+  kind: javascript/es5
+  file: ./index.js
+```
+
+### Mutations Schema
+
+The mutations schema defines all of the mutations in our subgraph. The mutations schema is a super-set of the subgraph's schema. For example, starting from a base subgraph schema:
```
Jannis: our -> the

Jannis: How about `The mutations schema is a super-set of the subgraph's schema` -> `The mutations schema builds on the subgraph schema by allowing to use types from the subgraph schema, as well as defining new types that are used only in the context of mutations.`?

Jannis: Because it's not a superset.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+  oldName: String!
+  newName: String!
+}
+
+type Mutation {
+  createEntity(
+    options: MyEntityOptions!
+  ): MyEntity!
+
+  setEnityName(
```
Jannis: Typo: `setEnityName` -> `setEntityName`

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+    entity: MyEntity!
+    name: String!
+  ): NewNameSet!
+}
+```
+
+`graph-cli` handles the combining, parsing, and validating of these two schemas. The `graph-cli` verifies that the mutations schema defines a `type Mutation`, that all of the mutations within it are defined in the resolvers module (see next section).
```
Jannis: It doesn't really `combine` them, let's just say `parsing and validating`

Jannis: And change ``` The `graph-cli` ``` to `It` 😄

Jannis: And change `, that all of the mutations`  to ` and that all of the mutations`

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+    name: String!
+  ): NewNameSet!
+}
+```
+
+`graph-cli` handles the combining, parsing, and validating of these two schemas. The `graph-cli` verifies that the mutations schema defines a `type Mutation`, that all of the mutations within it are defined in the resolvers module (see next section).  
+
+### Mutation Resolvers
+
+Each mutation within the schema must have a corresponding resolver function defined. Resolvers will be invoked by whatever engine executes the query. They are executed locally within the client application.
```
Jannis: `executes the query` - `executes the mutation queries`

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+
+Mutation resolvers of kind `javascript` take the form of a javascript module. This module is expected to have a default export that contains the following properties:
```
Jannis: `javascript` -> `javascript/es5` (for now)

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+  * resolvers - The mutation resolver functions.
+  * config - A collection of config generators.
```
Jannis: Let's mark resolvers and config as inline code with backticks: ``` `resolvers` ``` and ``` `config` ```

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+      const { name, value } = args.options
+
+      // Use config properties created by the
+      // config generator functions
+      const { ethereum, ipfs } = context.graph.config
+
+      // Fetch datasource addresses & abis
+      const { MyContract } = context.graph.datasources
+      await MyContract.abi
+      await MyContract.address
```
Jannis: This is giving me headaches right now. Can we leave out access to the data sources from the initial RFC? I need to think more about the interaction between the mutations and query engine / subgraph metadata.

Jannis: We may even be able to leave this out entirely. The mutation resolvers have to have built-in knowledge about which ABIs they interact with, and the contract addresses could be passed in as arguments.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+
+      // Modify a state object, which relays updates back
+      // to the subscribed dApp
+      const { state } = context.graph
+      state.addTransaction("tx_hash")
```
Jannis: This does not reflect the latest state API yet.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+}
+```
+
+### dApp Integration
+```javascript
+const {
+  createMutations,
+  createMutationsLink,
+  useMutation
+} = require("@graphprotocol/mutations-ts")
```
Jannis: This doesn't reflect the package split we discussed yet.

rfcs/0003-mutations.md
Jannis: Yes, I think that works. Can we leave out the `subgraph` and `node` options from the design for now (see my earlier description about the data sources)?

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,317 @@
...
+    },
+    ipfs: "http://localhost:5001",
+    customProperty: 5,
+    rootProperty: {
+      nestedProperty: "foo"
+    }
+  }
+})
+
+// Create an Apollo Links
```
Jannis: `an Apollo Links`? 😁

rfcs/0003-mutations.md
Jannis: This still needs to be updated

rfcs/0003-mutations.md
Jannis: This too

Jannis (changes requested):

rfcs/0003-mutations.md
dOrgJelli: Assuming the answer is yes here, as it's apart of the [manifest-schema](https://github.com/graphprotocol/graph-cli/blob/master/manifest-schema.graphql).

rfcs/0003-mutations.md
dOrgJelli: I was thinking we would just omit the specVersion within the mutations.yaml (if it's separate), as it's meaningless without the subgraph.yaml. If specVersion is only within subgraph.yaml, a mismatch would never occur.

rfcs/0003-mutations.md
dOrgJelli: Sure thing, would love to know more about what's erking you about this.

IMO this is an extremely valuable feature for mutations, as managing your contract addresses is a hassle when developing protocols & dApps. Having the subgraph be the source of truth would be wonderful.

rfcs/0003-mutations.md
dOrgJelli: 1. I've added type definitions for all API usages present within the example. Let me know if they're too verbose or if they should be detailed elsewhere in the RFC.
2. I've added the mutation state API to the RFC.

rfcs/0003-mutations.md
dOrgJelli: After talking with @Jannis about this we've decided to keep this in the RFC. As the Query Engine becomes production ready, the `createMutations` interface can be changed to support it.

rfcs/0003-mutations.md
dOrgJelli: Thanks so much for this link, it was able to help me find a configuration that worked.

I've run into one last problem: building for a node environment and the browser requires different "target" types in the webpack config. I'm looking into solutions now, but I have a feeling we may end up needing to support something along the lines of:
```yaml
resolvers:
  - kind: javascript/es5/node
    file: ./bundle-node/index.js
  - kind: javascript/es5/web
    file: ./bundle-web/index.js
```

dOrgJelli: This is the error I receive when trying to use a "node bundle" within the browser:
image

This is because...
> When running the code in the browser, there is no builtin require function. If you want to flag a module as external in the browser, you have to provide some other way to load it (put a script tag, or integrate an AMD library like require.js). Otherwise this external module cannot be imported.  

Taken from this response: https://github.com/liady/webpack-node-externals/issues/17#issuecomment-284222729

Jannis: I'm not sure why there even is a need to import anything. It's a bundle after all. I think I'll need some hands-on time with the bundling to figure something out.

rfcs/0003-mutations.md
Jannis: That's true, although if multiple subgraphs use the same mutations manifest, and one of them uses a different specVersion, we'd want to know if there is a mismatch. I think it's worth keeping.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Contents
+
+<!-- toc -->
+
+## Summary
+
+GraphQL mutations allow developers to add executable functions to their schema. Callers can invoke these functions using GraphQL queries. An introduction to how mutations are defined and work can be found [here](https://graphql.org/learn/queries/#mutations). This RFC will assume the reader understands how to use GraphQL mutations in a traditional Web2 application. This proposal describes how mutations are added to The Graph's toolchain, and used to replace web3 write operations the same way The Graph has replaced Web3 read operations.
```
Jannis: Web2, web3, Web3. One of these does not fit the pattern. 😁 Check the rest of the file as well.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+* _Mutation Resolvers_: Code module that contains all resolvers.  
+* _Resolver_: Function that is used to execute a mutation's logic.  
+* _Mutation Context_: A context object that's created for every mutation that's executed. It's passed as the 3rd argument to the resolver function.  
+* _Mutation States_: A collection of mutation states. One is created for each mutation being executed in a given query.  
+* _Mutation State_: The state of a mutation being executed. Also referred to in this document as "_State_". It is an aggregate of the core & ext states (see below). dApp developers can subscribe to the mutation's state upon execution of the mutation query. See the `useMutation` examples below.  
+* _Core State_: Default properties present within every mutation state. Some examples: `events: Event[]`, `uuid: string`, and `progress: number`.  
+* _Ext State_: Properties the mutation developer defines. These are added alongside the core state properties in the mutation state. There are no bounds to what a developer can define here. See examples below.
```
Jannis: Let's use complete words everywhere and call `ext state` either `extension state` or `extended state`. This includes type names etc.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+* _State Events_: Events emitted by mutation resolvers. Also referred to in this document as "_Events_". Events are defined by a `name: string` and a `payload: any`. These events, once emitted, are given to reducer functions which then update the state accordingly.  
+* _Core Events_: Default events available to all mutations. Some examples: `PROGRESS_UPDATE`, `TRANSACTION_CREATED`, `TRANSACTION_COMPLETED`.  
+* _Ext Events_: Events the mutation developer defines. See examples below.  
+* _State Reducers_: A collection of state reducer functions.  
+* _State Reducer_: Reducers are responsible for translating events into state updates. They take the form of a function that has the inputs [event, current state], and returns the new state post-event. Also referred to in this document as "_Reducer(s)_".  
+* _Core Reducers_: Default reducers that handle the processing of the core events.  
+* _Ext Reducers_: Reducers the mutation developer defines. These reducers can be defined for any event, core or ext. The core & ext reducers are run one after another if both are defined for a given core event. See examples below.
```
Jannis: Same: Spell out "Ext"

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+* _Core Reducers_: Default reducers that handle the processing of the core events.  
+* _Ext Reducers_: Reducers the mutation developer defines. These reducers can be defined for any event, core or ext. The core & ext reducers are run one after another if both are defined for a given core event. See examples below.  
+* _State Updater_: The state updater object is used by the resolvers to dispatch events. It's passed to the resolvers through the mutation context like so: `context.graph.state`.
+* _State Builder_: An object responsible for (1) initializing the state with initial values and (2) defining reducers for events.  
+* _Core State Builder_: A state builder that's defined by default. It's responsible for initializing the core state properties, and processing the core events with its reducers.  
+* _Ext State Builder_: A state builder defined by the mutation developer. It's responsible for initializing the ext state properties, and processing the ext events with its reducers.
```
Jannis: Here as well (Ext). Worth checking across the entire file.

rfcs/0003-mutations.md
Jannis: Or adding, rather.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+specVersion: ...
+...
+mutations:
+  file: ./mutations/mutations.yaml
+dataSources: ...
+...
+```
+`mutations/mutations.yaml`
+```yaml
+repository: https://npmjs.com/package/...
```
Jannis, line 106: Add `specVersion` 😉

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+schema:
+  file: ./schema.graphql
+resolvers:
+  apiVersion: 0.0.1
+  kind: javascript/es5
+  file: ./index.js
+  types: ./index.d.ts
+```
+
+NOTE: `resolvers.types` is optional, and is only required if the resolvers' module exports a state builder. More on this below.
```
Jannis: resolvers' -> resolvers

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+            name
+            value
+          }
+        }`
+      })
+
+      // Fetch datasource addresses & abis from the subgraph manifest
+      const { MyContract } = context.graph.datasources
+      await MyContract.abi
+      await MyContract.address
```
Jannis: This one we decided to leave out for now IIRC.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+  State,
+  EventMap,
+  MyEvent
+}
+```
+
+NOTE: If the resolvers module exports a `stateBuilder`, it's expected that the mutations manifest has a `resolvers.types` file defined. The following types are expected to be defined in the .d.ts type definition file:
+  - `State`
+  - `EventMap`
+  - Any `EventPayload` interfaces defined within the `EventMap`
```
Jannis: Isn't it still optional if the consumer is purely written in JS? We could say that it's strongly recommended.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+      mutation: DocumentNode
+      children: (
+        mutateFunction: MutationFunction<TData, TVariables>,
+        result: MutationResult<TData>
+      ) => JSX.Element | null
+    }
+
+    const Mutation = <TData = any, TVariables = OperationVariables>(
+      props: MutationComponentOptions<TData, TVariables>
+    ): JSX.Element | null => { ... }
```
Jannis: Is any of this specific to us / customized? If not, I think we can drop it.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+
+The existing alternative that protocol developers are creating for dApp developers has been described above.
+
+## Open Questions
+
+- **How can mutations pickup where they left off in the event of an abrupt application shutdown?**
+  Since mutations can contain many different steps internally, it would be ideal to be able to support continuing resolver execution in the event the dApp abruptly shuts down.
+
+- **How can dApps understand what steps a given mutation will take during the course of its execution?**
+  dApps may way to present to the user friendly progress updates, letting them know a given action is 3/4 of the way through execution (for example) and a high level description of each step. I view this as closely tied to the previous open question above, as we could support continuing resolver executions if we know what step it's currently undergoing. More research will be done, and a potential solution to solve both of these use cases will be presented at a later date.
```
Jannis: It's separate from the concern of how a mutation can be "resumed".

I agree this is an improvement we can make in the future. I have an API in mind for this already though, where the state has a `steps: Step[]` field, where each step has the following shape:
```typescript
interface Step {
  id: string
  title: string
  description: string
  status: 'pending' | 'processing' | 'error' | 'finished',
  current: boolean
  error: Error?
  data: any
}
```
That, plus a few core events would be all we need to render UIs like those on https://ant.design/components/steps/.

But let's revisit this out of band and outside this RFC.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+
+- **Should dApps be able to define event handlers for mutation events?**
+  dApps may want to implement their own handlers for specific events emitted from mutations. These handlers would be different form the reducers, as we wouldn't want them to be able to modify the state. Instead they could store their own state elsewhere within the dApp based on the events.
```
Jannis: That's a neat idea. Let's keep the open question here but not implement this until a use case comes along that requires it.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+
+- **Should the resolvers module be ES5 compliant?**  
+  The prototype was originally developed under these conditions. ES5 compliance has since been abandoned as it has proven nearly impossible to successfully transpile all dependencies into a single monolithic module.
```
Jannis: It still is my hope that this is possible. But I can accept the RFC even with this question still being open.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+
+- **Should the Graph Node's schema introspection endpoint respond with the "full" schema, including the mutations' schema?**
+  Developers could fetch the "full" schema by looking up the subgraph's manifest, read the `mutations.schema.file` hash value, and fetching the full schema from IPFS. Should the graph-node support querying this full schema directly from the graph-node itself through the introspection endpoint?
```
Jannis, line 665: It doesn't make sense to me, since graph-node doesn't provide mutation resolvers. So it would be including things in the introspection schema that it can't satisfy. We can leave the open question though.

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+  dApps may want to implement their own handlers for specific events emitted from mutations. These handlers would be different form the reducers, as we wouldn't want them to be able to modify the state. Instead they could store their own state elsewhere within the dApp based on the events.
+
+- **Should the resolvers module be ES5 compliant?**  
+  The prototype was originally developed under these conditions. ES5 compliance has since been abandoned as it has proven nearly impossible to successfully transpile all dependencies into a single monolithic module.
+
+- **Should the Graph Node's schema introspection endpoint respond with the "full" schema, including the mutations' schema?**
+  Developers could fetch the "full" schema by looking up the subgraph's manifest, read the `mutations.schema.file` hash value, and fetching the full schema from IPFS. Should the graph-node support querying this full schema directly from the graph-node itself through the introspection endpoint?  
+
+- **Will server side execution ever be a reality?**
+  I have not thought of a trustless solution to this, am curious if anyone has any ideas of how we could make this possible.
```
Jannis, line 668: Let's keep the open question for now :)

rfcs/0003-mutations.md
```diff
@@ -0,0 +1,619 @@
...
+- **Should the resolvers module be ES5 compliant?**  
+  The prototype was originally developed under these conditions. ES5 compliance has since been abandoned as it has proven nearly impossible to successfully transpile all dependencies into a single monolithic module.
+
+- **Should the Graph Node's schema introspection endpoint respond with the "full" schema, including the mutations' schema?**
+  Developers could fetch the "full" schema by looking up the subgraph's manifest, read the `mutations.schema.file` hash value, and fetching the full schema from IPFS. Should the graph-node support querying this full schema directly from the graph-node itself through the introspection endpoint?  
+
+- **Will server side execution ever be a reality?**
+  I have not thought of a trustless solution to this, am curious if anyone has any ideas of how we could make this possible.  
+
+- **Will The Graph Explorer support mutations?**
```
Jannis, line 670: Yeah, let's keep the open question.

Jannis (changes requested):
Looking good! Most of my comments are minor.

rfcs/0003-mutations.md
dOrgJelli: Ahh that's very true, will investigate further.

rfcs/0003-mutations.md
dOrgJelli: Okay sounds like a plan :+1:

rfcs/0003-mutations.md
dOrgJelli: I was able to successfully transpile the full bundle to ES5! Removing this.

Fun Fact: If you set `mode: "production"` in your webpack config, the generated bundle is no longer ES5 compliant... not sure why this is... Was able to get ES5 compliance by setting `mode: "development"`.

rfcs/0003-mutations.md
dOrgJelli: Changed `ext` to `extended`.

rfcs/0003-mutations.md
dOrgJelli: Correct, will take this out.

rfcs/0003-mutations.md
dOrgJelli: My personal opinion is that the developer should still be expected to write the interfaces manually within the .d.ts file. This way dApps will know what properties are within the state.

Deferring to you to make the final call here.

rfcs/0003-mutations.md
dOrgJelli: I've updated these type definitions, adding the mutation state. These types are now specific to this specification and different from the `react-apollo` package.

rfcs/0003-mutations.md
dOrgJelli: Re "It's seperate...": Interesting, would love to hear your thoughts on how resuming a mutation could be implemented.

Re "I have an API in mind...": Awesome I really like this! Would love to implement this in the near future.

Should I remove this open question?

dOrgJelli: I've added your suggested implementation to this open question.

rfcs/0003-mutations.md
dOrgJelli: @Jannis I've changed this to state that it is required to have a resolvers.types file. Additionally I've added the `Config` type to list of require exported types.

This is because, when executing a mutation without Apollo, the developer needs to define a context for the execution to use. This `MutationContext` type requires type parameters for `Config`, `State`, and `EventMap`. An example of this can be seen here: https://github.com/dOrgTech/the-graph-mutations-spec/blob/586018191737a7606957c49cecd081021ac7abd1/packages/mutations/example/dapp.ts#L29-L52

Jannis: Ok, let's leave it as mandatory for now then.

rfcs/0003-mutations.md
Jannis: Great, thanks.

Re: resuming mutations. I think the best approach here would be to force resolvers into implementing a state machine that has a `restore` method or similar that takes a previously persisted state and identifies the state it should be in.

But for now I'd rather not force mutation resolvers into such a strict pattern. It could be a future add-on.

rfcs/0003-mutations.md
Jannis: Heh. Nena may have some ideas here, but let's figure this out in the coming weeks.

Jannis (approved):
Alright, here we go!

PR #11, opened 2020-01-06
RFC-0004: Fulltext Search

fordN:
A proposal for a new feature to allow subgraph developers to add fulltext search indexes to their entities and thus expose fulltext search ranked filtering in the GraphQL API.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -6,6 +6,7 @@
     - [RFC-0001: Subgraph Composition](./rfcs/0001-subgraph-composition.md)
     - [RFC-0002: Ethereum Tracing Cache](./rfcs/0002-ethereum-tracing-cache.md)
     - [RFC-0003: Mutations](./rfcs/0003-mutations.md)
+    - [RFC-0004: Fulltext Search](./rfcs/0002-fulltext-search.md)
   - [Obsolete RFCs](./rfcs/obsolete.md)
   - [Rejected RFCs](./rfcs/rejected.md)
 - [Engineering Plans](./engineering-plans/index.md)
--- a/rfcs/0004-fulltext-search.md
+++ b/rfcs/0004-fulltext-search.md
@@ -91,15 +91,17 @@ For each fulltext search index a subgraph developer must be able to specify:
 The proposed process of adding a fulltext search index to an Entity 
 involves adding a special field to that Entity in the GraphQL schema of 
 the subgraph's definition. The subgraph developer will create a new virtual 
-field of type String that is marked with the special `@FullTextSearch` directive. 
-The directive will have two required parameters for specifying the language
- and the fields to be incorporated. Verification of the index definition 
- will ensure that all fields referenced are valid String type fields on 
- that entity. To be clear, the proposed interface will add the fulltext 
- search definitions to the subgraph deployment hash, so adding or updating 
- a fulltext search index will require an update to the subgraph.  
- With subgraph composition it will be possible to easily create new subgraphs 
- that add specific fulltext search indexes to an existing subgraph. 
+field of type String that is marked with the special `@index` directive. 
+The directive will have two required parameters for specifying the index
+`type` and `params`. Params fields are specific to the index type; in the 
+case of fulltext search there will be two params to specify: `language` 
+and `fields`. Verification of the index definition will ensure that all 
+fields referenced are valid String type fields. To be clear, the proposed 
+interface will add the fulltext search index definitions to the subgraph 
+deployment hash, so adding or updating a fulltext search index will require 
+an update to the subgraph.  With subgraph composition it will be possible 
+to easily create new subgraphs that add specific fulltext search indexes 
+to an existing subgraph. 
 
 Example fulltext search field definition:
 ```graphql
@@ -110,7 +112,17 @@ type ExampleItem @entity {
     specs: String!,
     address: Bytes!,
     marked: Boolean!,
-    itemSearch: String! @fullTextSearchIndex(language: "english", fields: [{name: "name", weight: 5}, {name: "description", weight: 3}, {name: "specs", weight: 1}])
+    itemSearch: String! @index(
+        type: "fulltext", 
+        params: {
+            language:"english", 
+            fields: [
+                {name: "name", weight: 5}, 
+                {name: "description", weight: 3}, 
+                {name: "specs", weight: 1},                
+            ],            
+        }
+    )
 }
 ```  
 
@@ -120,9 +132,9 @@ For each fulltext index three search algorithms will be made available to
 the end user: a simple text query, a standard ranking, and a cover 
 density ranking algorithm: 
   - `query`: search for the string in the indexed document. Several 
-    operators available: and, or, and proximity (`&`, `|`, `<->`.) 
+    operators are available: and, or, and proximity (`&`, `|`, `<->`.) 
     The proximity operator allows one to specify max distance: 
-    3 words apart → `<3>`. (filter suffix = ``)
+    3 words apart → `<3>`. (no filter suffix)
   - `standard ranking`: ranking based on the number of matching lexemes. 
     (filter suffix = `_rank`)
   - `cover density ranking`: Cover density is similar to the standard 
@@ -130,12 +142,12 @@ density ranking algorithm:
     to each other is taken into consideration. This function requires 
     lexeme positional information to perform its calculation, so it ignores 
     any "stripped" lexemes in the index.
-    (filter suffix = `_rank_cd`)
+    (filter suffix = `_proximity_rank`)
     
 Example Usage in a query (references the example Entity definition from above section): 
 ```graphql
 query {
-    exampleItems(first: 10, where: {itemSearch_rank_cd: "sarcastic"}) {
+    exampleItems(first: 10, where: {itemSearch_proximity_rank: "sarcastic&tea"}) {
         address
         name 
         description
@@ -147,7 +159,7 @@ query {
 
 Fulltext search query system implementations often involve specific systems 
 for storing and querying the text documents; however, in an effort to reduce 
-system complexity and feature implemtation time I propose starting with the 
+system complexity and feature implementation time I propose starting with the 
 fulltext search features built in to PostgreSQL.
 
 A FullText search field will get its own column in the Entity table just 
@@ -162,17 +174,17 @@ the indexed fulltext search data.
  
 
 Limitations of Postgres fulltext search indexes and search algorithms: 
-  - only certain languages available (expandable with plugins namely PGroonga), 
+  - only certain languages available (expandable with plugins like [PGroonga](https://pgroonga.github.io/)), 
   - select algorithms are available, and 
   - limited to PostgreSQL storage. 
 
 Alternative technologies (not in any particular order): 
-  Written in Rust:
-    - Tantivy (https://github.com/tantivy-search/tantivy)
-    - Toshi (https://github.com/toshi-search/Toshi)
-    - Sonic (https://github.com/valeriansaliou/sonic)
-    - MeiliSearch (https://github.com/meilisearch/MeiliSearch)
-    - Bayard (https://github.com/bayard-search/bayard)
+  - Written in Rust:
+    - [Tantivy](https://github.com/tantivy-search/tantivy)
+    - [Toshi](https://github.com/toshi-search/Toshi)
+    - [Sonic](https://github.com/valeriansaliou/sonic)
+    - [MeiliSearch](https://github.com/meilisearch/MeiliSearch)
+    - [Bayard](https://github.com/bayard-search/bayard)
 
 ## Compatibility
 
@@ -218,7 +230,7 @@ verification and allowing indexer node operators to experiment with
 algorithms and indexes in order to continue to improve query speed and results? 
 
 Since a fulltext search field is purely derivative of other Entity data 
-the addition or update of a @fullTextSearchIndex field does not require a 
-full blockchain resync, rather the index itself just needs to be rebuilt. 
+the addition or update of an @index field does not require a full blockchain 
+resync, rather the index itself just needs to be rebuilt. 
 How can we allow fulltext search index (and eventually other index types) 
 updates without requiring a full subgraph resync? 
--- a/rfcs/0004-fulltext-search.md
+++ b/rfcs/0004-fulltext-search.md
@@ -1,4 +1,4 @@
-# RFC-0000: Template
+# RFC-0004: Fulltext Search
 
 <dl>
   <dt>Author</dt>
@@ -26,40 +26,36 @@
 
 ## Summary
 
-This RFC proposes adding a fulltext search filter type to the GraphQL API 
-to allow subgraph developers to specify language-specific, lexical, 
-composite filters that end users can use in their queries. The fulltext 
-index will be created by examining all words in a document, breaking it 
-into individul words and phrases (lexical analysis), and collapsing 
+The fulltext search filter type is a feature of the GraphQL API that
+allows subgraph developers to specify language-specific, lexical,
+composite filters that end users can use in their queries. The fulltext
+search feature examines all words in a document, breaking it into
+individual words and phrases (lexical analysis), and collapsing
 variations of words into a single index term (stemming.)
 
 ## Goals & Motivation
 
-The current set of string filters available in our GraphQL API is lacking 
-the tools needed to build effective, modern interfaces. Wildcard string 
-matching does provide string filtering, but users have come to expect the 
-easy to use filtering that comes with fulltext search systems.
+The current set of string filters available in the GraphQL API is lacking 
+fulltext search capabilities that enable efficient searches across entities
+and attributes. Wildcard string matching does provide string filtering, but 
+users have come to expect the easy to use filtering that comes with fulltext 
+search systems.
 
-To facilitate building effective user interfaces we must provide human-user 
-friendly query filtering. Lexical, composite fulltext search filters can 
-provide the tools necessary for front-end developers to implement powerful 
-search bars that filter data across multiple fields of an Entity.
+To facilitate building effective user interfacee human-user friendly query 
+filtering is essential. Lexical, composite fulltext search filters can provide 
+the tools necessary for front-end developers to implement powerful search 
+bars that filter data across multiple fields of an Entity.
 
-With the proposed feature I aim to:
-  - provide tools for subgraph developers to define composite search 
-    indexes that can search across multiple fields and entities, and
-  - integrate powerful, user friendly text search that is flexible and 
-    can be easily defined by a subgraph developer ultimately providing 
-    the tools necesssary for front end developers to create effective 
-    user interfaces without the need to middleware.
+The proposed feature aims to provide tools for subgraph developers to define 
+composite search APIs that can search across multiple fields and entities.
      
 ## Urgency
 
-A delay in adding the fulltext search feature  will not create issues with 
-current deployments. However, it will represent a realization of part of 
-the long term vision for the query network. Also several high profile users 
-have communicated that it may be a conversion blocker, so implementation 
-should be prioritized. 
+A delay in adding the fulltext search feature will not create issues
+with current deployments. However, the feature will represent a
+realization of part of the long term vision for the query network. In
+addition, several high profile users have communicated that it may be a
+conversion blocker. Implementation should be prioritized. 
 
 ## Terminology
 
@@ -86,39 +82,35 @@ should be prioritized.
 
 ## Detailed Design
 
-### Subgraph Definition
+### Subgraph Schema
 
-Part of the power of fulltext search indexes are their flexibility, so 
-it is important to expose a simple interface to facilitate useful 
-applications of the index and ulimately aim to reduce the need to create 
-new subgraphs for the express purpose of updating fulltext search fields 
-for end-use cases. 
+Part of the power of the fulltext search API is the flexibility, so 
+it is important to expose a simple interface to facilitate useful applications 
+of the index and aim to reduce the need to create new subgraphs for the 
+express purpose of updating fulltext search fields. 
 
-For each fulltext search index a subgraph developer must be able to specify:
+For each fulltext search API a subgraph developer must be able to specify:
     1. a language, 
     2. a set of text document fields to include,
-    3. relative weighting for each field in the index.
-    4. a choice of ranking algorithm for sorting query result items
+    3. relative weighting for each field,
+    4. a choice of ranking algorithm for sorting query result items.
 
-The proposed process of adding one or more fulltext search index to an Entity 
-involves adding one or more fulltext directive to the `_Schema_` type in the 
-subgraph's GraphQL schema.  Each fulltext definition will have four required
-top level parameters: `name`, `language`, `algorithm`, and `include`. 
-The fulltext search definitions will be used to generate queries on the GraphQL
-schema that will be exposed to the end user.  
+The proposed process of adding one or more fulltext search API involves
+adding one or more fulltext directive to the `_Schema_` type in the
+subgraph's GraphQL schema. Each fulltext definition will have four
+required top level parameters: `name`, `language`, `algorithm`, and
+`include`. The fulltext search definitions will be used to generate
+query fields on the GraphQL schema that will be exposed to the end user.
 
-Applying fulltext search indexes across entities will be a powerful 
-abstraction allowing users to search across all relevant entities in one
-query. An index search across entities will by definition have polymorphic
-results, so in that case a union type will also be generated in the schema 
-for the query result items. 
+Enabling fulltext search across entities will be a powerful abstraction 
+that allows users to search across all relevant entities in one query. Such 
+a search will by definition have polymorphic results. To address this, a 
+union type will be generated in the schema for the fulltext search results. 
 
-Verification of the index definition will ensure that all fields referenced 
-are valid String type fields. To be clear, the proposed interface will add 
-the fulltext search index definitions to the subgraph deployment hash, so 
-adding or updating a fulltext search index will require an update to the subgraph. 
-With subgraph composition it will be possible to easily create new subgraphs 
-that add specific fulltext search indexes to an existing subgraph. 
+Validation of the fulltext definition will ensure that all fields referenced 
+in the directive are valid String type fields. With subgraph composition 
+it will be possible to easily create new subgraphs that add specific fulltext 
+search capabilities to an existing subgraph. 
 
 Example fulltext search definition:
 
@@ -131,7 +123,7 @@ type _Schema_
   @fulltext(
     name: "search",
     language: "english",
-    algorithm: "ranked",
+    algorithm: ranked, # variant of `_FullTextAlgorithm` enum
     include: [
       {
         entity: "Band",
@@ -158,28 +150,32 @@ type _Schema_
 
 The schema generated from the above definition: 
 ```graphql
-union FulltextMediaResultItem = ...
-union FulltextSearchResultItem = Band | Album | Musician
+union _FulltextMediaEntity = ...
+union _FulltextSearchEntity = Band | Album | Musician
 type Query {
   media...
-  search(text: String!, first: Int, skip: Int, block: Int): [FulltextSearchResultItem!]!
+  search(text: String!, first: Int, skip: Int, block: Block_height): [FulltextSearchResultItem!]!
 }
 ```
 
 ### GraphQL Query interface
 
-End users of the subgraph will have access to the queries generated for the
-fulltext search indexes alongside the other queries generated for each 
-entity in the subgraph.  In the case of a fulltext search index that spans
-multiple entities, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) 
-may be used in the query to deal with the polymorphic result items. In the 
-front-end the `__typename` in each result item can be used to render the 
-components accordingly.  
+End users of the subgraph will have access to the fulltext search
+queries alongside the other queries available for each entity in the
+subgraph. In the case of a fulltext search defined across multiple
+entities,
+[inline fragments](https://graphql.org/learn/queries/#inline-fragments)
+may be used in the query to deal with the polymorphic result items. In
+the front-end the `__typename` field can be used to distinguish the
+concrete entity types of the returned results.
 
-In the "text" parameter supplied to the query there will be one special 
-operator supported, the proximity operator. Several operators are available: 
-and, or, and proximity (`&`, `|`, `<->`.) The proximity operator allows 
-one to specify max distance between search terms: 3 words apart → `<3>`. 
+In the `text` parameter supplied to the query there will be several operators
+available to the end user.  Included are the and, or, and proximity operators
+(`&`, `|`, `<->`.) The special, proximity operator allows clients to specify 
+the maximum distance between search terms: `foo<3>bar` is equivalent to 
+requesting that `foo` and `bar` are at most three words apart. 
+
+
 
 Example query using inline fragments and the proximity operator: 
 ```graphql
@@ -197,33 +193,18 @@ query {
 
 Fulltext search query system implementations often involve specific systems 
 for storing and querying the text documents; however, in an effort to reduce 
-system complexity and feature implementation time I propose starting with the 
-fulltext search features built in to PostgreSQL.
+system complexity and feature implementation time I propose starting with 
+extending the current store interface and storage implemenation with fulltext 
+search features rather than use a fulltext specific interface and storage
+system.
 
-A FullText search field will get its own column in the Entity table just 
-like any other field; however, the data stored will be the result of the 
-lexical, morphological analysis of text documents performed on the fields 
-included in the index. The fulltext search field will be created using the 
-Postgres ts_vector function and will be indexed using a GIN index. The subgraph 
-developer will define a ranking algorithm to be used to sort query results,
-so the end-user facing API remains easy to use without any requirement to 
-understand the ranking algorithms.   
-
-Limitations of Postgres fulltext search indexes and search algorithms: 
-  - only certain languages available (expandable with plugins like [PGroonga](https://pgroonga.github.io/)), 
-    - **languages supported out of the box**: 
-        Danish, Dutch, English, Finnish, French, German, Hungarian, Italian, 
-        Norwegian, Portuguese, Romanian, Russian, Spanish, Swedish, and Turkish.
-  - select algorithms are available, and 
-  - limited to PostgreSQL storage. 
-
-Alternative technologies (not in any particular order): 
-  - Written in Rust:
-    - [Tantivy](https://github.com/tantivy-search/tantivy)
-    - [Toshi](https://github.com/toshi-search/Toshi)
-    - [Sonic](https://github.com/valeriansaliou/sonic)
-    - [MeiliSearch](https://github.com/meilisearch/MeiliSearch)
-    - [Bayard](https://github.com/bayard-search/bayard)
+A FullText search field will get its own column in a table dedicated to fulltext
+data. The data stored will be the result of the lexical, morphological analysis 
+of text documents performed on the fields included in the index. The fulltext 
+search field will be created using the Postgres ts_vector function and will 
+be indexed using a GIN index. The subgraph developer will define a ranking 
+algorithm to be used to sort query results,so the end-user facing API remains 
+easy to use without any requirement to understand the ranking algorithms.   
 
 ## Compatibility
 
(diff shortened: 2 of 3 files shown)


rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,223 @@
...
+Example fulltext search field definition:
+```graphql
+type ExampleItem @entity {
+    id: ID!,
+    name: String!,
+    description: String!,
+    specs: String!,
+    address: Bytes!,
+    marked: Boolean!,
+    itemSearch: String! @fullTextSearchIndex(language: "english", fields: [{name: "name", weight: 5}, {name: "description", weight: 3}, {name: "specs", weight: 1}])
```
fordN: Some input on the directive name is welcome here! 
With the possibility in mind of there being additional custom indexes later,  an option I've considered is using a more general directive such as: `@index(type: String, param1: [String], param2: [String])`

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,223 @@
...
+  - provide tools for subgraph developers to define composite search 
+    indexes that search across multiple fields in an entity, and
+  - integrate powerful, user friendly text search that is flexible and 
+    can be easily defined by a subgraph developer ultimately providing 
+    the tools necesssary for front end developers to create effective 
+    user interfaces without the need to middleware.
+     
+## Urgency
+
+Fulltext search filters are not essential to the value proposition of
```
yanivtal: I don't see how it's true that fulltext search isn't essential to the value proposition of Graph Node. Disagree!

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,223 @@
...
+matching does provide string filtering, but users have come to expect the 
+easy to use filtering that comes with fulltext search systems.
+
+To facilitate building effective user interfaces we must provide human-user 
+friendly query filtering. Lexical, composite fulltext search filters can 
+provide the tools necessary for front-end developers to implement powerful 
+search bars that filter data across multiple fields of an Entity.
+
+With the proposed feature I aim to:
+  - provide tools for subgraph developers to define composite search
```
yanivtal: I think fulltext search indexes should be able to span across entities so this language is a little limiting.

rfcs/0004-fulltext-search.md
yanivtal: I thought we decided to move away from including this in the schema. When did we decide to put it back?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,223 @@
...
+    fulltext search ranking except that the proximity of matching lexemes 
+    to each other is taken into consideration. This function requires 
+    lexeme positional information to perform its calculation, so it ignores 
+    any "stripped" lexemes in the index.
+    (filter suffix = `_rank_cd`)
+    
+Example Usage in a query (references the example Entity definition from above section): 
+```graphql
+query {
+    exampleItems(first: 10, where: {itemSearch_rank_cd: "sarcastic"}) {
```
yanivtal: I dislike this API. It exposes a lot of complexity to the developer. The API should default to simple and ergonomic with more complex capabilities added as optional parameters.

rfcs/0004-fulltext-search.md
yanivtal: I much prefer the @index approach

yanivtal (changes requested):

rfcs/0004-fulltext-search.md
fordN: Yep, I have no argument there; fulltext search style querying is vital to the value proposition.  

I believe what I was trying to convey is that: if the feature is not added immediately the Graph Node still has a strong value proposition and most current and prospective users will still be happy.  Maybe I'm being too literal about the difference between urgency and priority; but it's going to be a valuable feature that will likely get a lot of use, let's get it added right away! :)

rfcs/0004-fulltext-search.md
fordN: When you say "move away from including this in the schema" I'm assuming you are also talking about the decision to include it in the subgraph definition in the schema? Or are you simply saying it should be in the subgraph definition, but somewhere else?

fordN: I have definitely gone back and forth on putting it in the base subgraph definition or allowing it to be an add on to an already deployed subgraph, but we had a brief discussion a few weeks ago on a call; Jannis and I (if I remember correctly) agreed that it needs to be a part of the definition of uniqueness for a subgraph to keep a stable GraphQL API for each subgraph. It won't work to have a query filter or type change out from under a user if indexes are updated. With the subgraph composition though we will be able to maintain flexibility for subgraph developers to add indexes to existing subgraphs.  

Looking forward to the decentralized network, the more I thought through the added complexity of essentially adding an indexing marketplace to the query marketplace the more it seems that it poses risks to the health of the data network. 
  - Will index updates have their own system for being propagated to Graph Nodes storing that subgraph? Why not integrate them into the definition of a subgraph, so we can leverage the existing query discovery and subgraph data propagation systems.
  -  A query engine will have to find a node that not only has the subgraph deployment, but also the correct combination of indexes applied to it.  This will involve propagating index updates to the network and could result in a less clear graph of data.  
- On the other hand you could argue that forcing users to create a new subgraph in order to update or add a fulltext search index will lead to a large set of very similar subgraphs, but I _think_ the complexity of organizing that through interface design will in the end be simpler than managing different schemas/indexes on a single subgraph.  I also think the aforementioned interface work will be useful or needed for managing the complexity introduced by subgraph composition which could also leading to situations where many similar subgraphs are being created.
- Perhaps less important than correctness on the decentralized network, but keep in mind there will need to be new interfaces designed and built on Graph Explorer for managing indexes.

Concerning the choice of putting the definition in the schema: I've also thought about creating yaml files for the index definition or adding to the subgraph definition, but the schema is where the data model and query model is defined and it makes sense to keep it query concerns there.  Though it might make sense to separate raw data model concerns from front end query concerns, we are already using @derivedFrom for virtual fields that are derived from Entity data and I think keeping the query model in one place together with the data model is a useful simplification.

yanivtal: Alright I see the case for this. Then can we look at these changes:
1. Switch to @index directive with a type: "fulltext"
2. Make sure the fields attributes can include fields on other entities in the schema (including imported types from subgraph composition)?

rfcs/0004-fulltext-search.md
fordN: I think full text search could be very useful across entities, so I've updated this language to include that intent. 
With that goal in mind I'm working on how to enable this behavior in the GraphQL API while still encouraging data models that will give performant querying and index building.

rfcs/0004-fulltext-search.md
fordN: I've updated the API to use more general, self-describing language: `rank_cd` -> `proximity_rank`.  Keeping in mind the query can be made without any suffixes do you think this reduces the complexity you pointed out?

There is also the complexity of using operators in the filter: `&`, `|` and `<->`.  I think these operators are general and clear enough, but their usage would need to be conveyed through documentation or schema descriptions as it may not be obvious.

rfcs/0004-fulltext-search.md
Zerim: I'm not a big fan of the directive approach and even less so when on top of a virtual attribute whose only purpose is to support the functioning of an index.

Indexes and data models are separate concerns, and forcing storage level concerns into the application data model hurts re-usability and limits flexibility for the Indexer.

I think I'd rather see this go somewhere in the manifest under i.e., `suggested_indexes`, where the subgraph developer could also provide hints to other indexes that they think may be useful initially and Indexers can choose to incorporate into their decisions until they have enough data on query patterns to make that decision for themselves.

Zerim: It seems like a lot of thought has gone into this and the concerns above around discovery and alignment around the right text full-text search indexes seem valid. These are definitely different from normal attribute indexes in that they are much harder to organically align on in a distributed fashion based on query access patterns (because you still need to choose what type of search index).

Even so, combining the data model for an entity w/ a specific way of searching that entity feels wrong to me... could we still include the same information in the subgraph, but somewhere else in the manifest, perhaps as a "hint" to the Indexer rather than a mandate?

Indexes provided as hints could be given a label or an ID, which end-users can then use to reference in their query to force Indexers to use a specific index in resolving a query. This concept could be extended beyond full-text search indexes as well.

fordN: We could define custom indexes in the subgraph manifest (yaml) rather than in the schema, but I think that makes it more difficult to manage. 
- A new index also means a new query with specific behavior, so I'm having trouble following how it would work as just "a hint to the Indexer".  
- We currently have many indexes per subgraph; they are derived from the schema.  Why change the paradigm?
- All of our query types are currently generated from information in the schema.  Why split query definitions across files in the subgraph?
- The index definition will consist mostly of entity field names to be included these are in the schema, so working within one file (while not a huge difference) will be easier. 

Rather than give indexes a unique id (which will not be self-describing) I think we should creates a unique query for the index, so end-users can use that specific query to force the Indexer to use a specific index. Indexers can implement that index using their own custom systems, but it should always search across the fields described in the schema (or manifest based on your suggestion). The query name can be defined in the manifest by the subgraph developer.

fordN: Given the need to apply full-text search indexes that span multiple entities, I would like to update the proposal to have the index defined as a separate type in the schema or an extension to the query type. This way the return type can be a `union` or `interface` type that includes the affected entities and a new query will be generated rather than simply adding a new filter to the existing entity query.  
Including the index definition in the manifest rather then the schema would work too, though I personally prefer the schema. See my reasoning in response to your other comment below.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,236 @@
...
+
+A delay in adding the fulltext search feature  will not create issues with 
+current deployments. However, it will represent a realization of part of 
+the long term vision for the query network. Also several high profile users 
+have communicated that it may be a conversion blocker, so implementation 
+should be prioritized. 
+
+## Terminology
+
+- `lexeme`: a basic lexical unit of a language, consisting of one word or
```
Jannis: We've been using italics (`/.../`) for special terms so far. Let's use them here too, so we maintain consistence across documents.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,236 @@
...
+Example fulltext search field definition:
+```graphql
+type ExampleItem @entity {
+    id: ID!,
+    name: String!,
+    description: String!,
+    specs: String!,
+    address: Bytes!,
+    marked: Boolean!,
+    itemSearch: String! @index(
```
Jannis: If fulltext search is expected to work across entities, then this feels wrong. One idea could be to define indexes globally on the `_Schema_` type that we're introducing for e.g. schema composition:
```graphql
type _Schema_
  @index(
    index: {
      name: "search",
      language: "english",
      include: [
        { entity: "ExampleItem", fields: [ {name: "name", weight: 5 }, ... ] }
      ]
    }
  )
```

Note: In GraphQL, there is no such thing as an any type, which is why it is better not to decouple `type` and `params` and instead define possible indexes with something like
```graphql
union IndexDefinition = FulltextIndexDefinition | OtherIndexDefinition

type FulltextIndexEntityField {
  name: String!
  weight: Int!
}

type FulltextIndexEntity {
  entity: String!
  fields: [FulltextIndexEntityField!]!
}

type FulltextIndexDefinition {
  name: String!
  language: String!
  entities: [FulltextIndexEntity!]!,
}
```

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,236 @@
...
+    
+Example Usage in a query (references the example Entity definition from above section): 
+```graphql
+query {
+    exampleItems(first: 10, where: {itemSearch_proximity_rank: "sarcastic&tea"}) {
+        address
+        name 
+        description
+    }
+}
```
Jannis: With my above suggestion, the schema-wide index name could be used as the query field. A fulltext search with the name `search` could be used as follows in queries:
```graphql
{
  ranked: search(first: 10, by: { rank: "sarcastic&tea" }, orderBy: rank, orderDirection: desc) {
    rank
    entity {
      ... on ExampleItem {
        address
        name
      }
    }
  }

  text: search(first: 10, by: { text: "foobar" }) {
    ...
    entity {
      ... on ExampleItem {
        address
        name
      }
    }
  }
}
```

One question I have is how much information the return values should carry in addition to the entity data.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+### Tools and Design
+
+Fulltext search query system implementations often involve specific systems 
+for storing and querying the text documents; however, in an effort to reduce 
+system complexity and feature implementation time I propose starting with the 
+fulltext search features built in to PostgreSQL.
```
Jannis: When we build query verification into the protocol, we may not be able to use Postgres for storing entity data anymore. If we build Postgres-specific fulltext search features into the protocol _now_, we may likely have to break the protocol later. @yanivtal @Zerim Thoughts?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,236 @@
...
+for indexes. This design moved the definition of custom indexes out of the 
+subgraph definition; the subgraph would be deployed without them and they 
+could be added later using a new Graph Explorer interface (in Hosted-Service context) 
+or a JSON-RPC request directly to a Graph Node. One of the benefits of this 
+design is that the definition of uniqueness for a subgraph does not include 
+the custom indexes, so a new subgraph deployment and corresponding syncing 
+work does not have to be added in order to create or update an index. 
+However, it also introduces significant added complexity: will a separate 
+query marketplace and discovery registry then be required for finding nodes 
+with your needed subgraph-index combination.
```
Jannis: I feel we should nonetheless include as little algorithmic details in what is included in the subgraph manifest (and related files). Primarily, what should be included is anything that is required to generate an API that includes fulltext search.

An approach different from declaring _indexes_ could be to declare special fulltext query fields:
```graphql
type _Schema_
  @fulltextSearchField(
    name: "search",
    entities: [ { name: "ExampleItem", fields: [ ... ] } ]
  )
```
or similar.

Jannis (changes requested):

rfcs/0004-fulltext-search.md
Zerim: Do we have any choice here? Are there full-text search features that are common and deterministic to all SQL databases.

For now, I'm OK w/ us having SQL or even Postgres specific full-text search functionality in the protocol w/ the understanding that there's a lot that may have to change when we eventually move to index chains and V2.

SUMMARY.md
```diff
@@ -5,6 +5,7 @@
   - [Approved RFCs](./rfcs/approved.md)
     - [RFC-0001: Subgraph Composition](./rfcs/0001-subgraph-composition.md)
     - [RFC-0002: Ethereum Tracing Cache](./rfcs/0002-ethereum-tracing-cache.md)
+    - [RFC-0004: Fulltext Search](./rfcs/0002-fulltext-search.md)
```
Jannis, line 9: Link and RFC number don't match.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
+# RFC-0000: Template
```
Jannis: This would be `RFC-0004: Fulltext Search`

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+## Contents
+
+<!-- toc -->
+
+## Summary
+
+This RFC proposes adding a fulltext search filter type to the GraphQL API 
+to allow subgraph developers to specify language-specific, lexical, 
+composite filters that end users can use in their queries. The fulltext 
+index will be created by examining all words in a document, breaking it
```
Jannis: That it's an index is an implementation detail. How about the `fulltext search feature will examine...` instead of `fulltext index will be created by examining...`?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+to allow subgraph developers to specify language-specific, lexical, 
+composite filters that end users can use in their queries. The fulltext 
+index will be created by examining all words in a document, breaking it 
+into individul words and phrases (lexical analysis), and collapsing 
+variations of words into a single index term (stemming.)
+
+## Goals & Motivation
+
+The current set of string filters available in our GraphQL API is lacking 
+the tools needed to build effective, modern interfaces. Wildcard string
```
Jannis: I don't agree with this statement. How about we say something like this instead?
> The current set of filters available in the GraphQL API is lacking fulltext search capabilities that enable searching for strings across entities and attributes efficiently. This is needed for search functionality in dApps.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+the tools needed to build effective, modern interfaces. Wildcard string 
+matching does provide string filtering, but users have come to expect the 
+easy to use filtering that comes with fulltext search systems.
+
+To facilitate building effective user interfaces we must provide human-user 
+friendly query filtering. Lexical, composite fulltext search filters can 
+provide the tools necessary for front-end developers to implement powerful 
+search bars that filter data across multiple fields of an Entity.
+
+With the proposed feature I aim to:
```
Jannis: `With the proposed feature I aim to` -> `The proposed feature aims to`

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+  - provide tools for subgraph developers to define composite search 
+    indexes that can search across multiple fields and entities, and
```
Jannis: search indexes -> search APIs
or
search indexes -> search fields

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+  - integrate powerful, user friendly text search that is flexible and 
+    can be easily defined by a subgraph developer ultimately providing 
+    the tools necesssary for front end developers to create effective 
+    user interfaces without the need to middleware.
```
Jannis: `the need to middleware` reads like `to middleware` is a verb. 😉 

The whole bullet point is vague and wordy for that. Perhaps we can drop it and turn the point above it into a regular paragraph.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+     
+## Urgency
+
+A delay in adding the fulltext search feature  will not create issues with 
+current deployments. However, it will represent a realization of part of 
+the long term vision for the query network. Also several high profile users 
+have communicated that it may be a conversion blocker, so implementation 
+should be prioritized.
```
Jannis: This reads as if a delay `will represent a realization of parrt of the long term vision`. Worth rephrasing.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+  are to a particular query, so that when there are many matches the most 
+  relevant ones can be shown first." [- Postgres Documentation](https://www.postgresql.org/docs/11/textsearch-controls.html#TEXTSEARCH-RANKING-SEARCH-RESULTS)
+  
+  **Algorithms**:
+  - _standard ranking_: ranking based on the number of matching lexemes.     
+  - _cover density ranking_: Cover density is similar to the standard 
+    fulltext search ranking except that the proximity of matching lexemes 
+    to each other is taken into consideration. This function requires 
+    lexeme positional information to perform its calculation, so it ignores 
+    any "stripped" lexemes in the index.
```
Jannis, line 81: Can we link to specifications of these algorithms?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+    fulltext search ranking except that the proximity of matching lexemes 
+    to each other is taken into consideration. This function requires 
+    lexeme positional information to perform its calculation, so it ignores 
+    any "stripped" lexemes in the index.            
+
+## Detailed Design
+
+### Subgraph Definition
+
+Part of the power of fulltext search indexes are their flexibility, so
```
Jannis: I still don't like that we're talking about `indexes` so much here. To me, they are nothing but an implementation detail. We should be talking purely about the fulltext search API and its behavior.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Subgraph Definition
+
+Part of the power of fulltext search indexes are their flexibility, so 
+it is important to expose a simple interface to facilitate useful 
+applications of the index and ulimately aim to reduce the need to create 
+new subgraphs for the express purpose of updating fulltext search fields 
+for end-use cases. 
+
+For each fulltext search index a subgraph developer must be able to specify:
```
Jannis: `index` -> API (or field)?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+    1. a language, 
+    2. a set of text document fields to include,
+    3. relative weighting for each field in the index.
```
Jannis: Drop `in the index.`?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+    4. a choice of ranking algorithm for sorting query result items
+
+The proposed process of adding one or more fulltext search index to an Entity
```
Jannis: `search index` -> `search API` (or field)?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+    1. a language, 
+    2. a set of text document fields to include,
+    3. relative weighting for each field in the index.
+    4. a choice of ranking algorithm for sorting query result items
+
+The proposed process of adding one or more fulltext search index to an Entity 
+involves adding one or more fulltext directive to the `_Schema_` type in the 
+subgraph's GraphQL schema.  Each fulltext definition will have four required
+top level parameters: `name`, `language`, `algorithm`, and `include`. 
+The fulltext search definitions will be used to generate queries on the GraphQL
```
Jannis: It will not `generate queries`, it'll generate query fields in the GraphQL schema.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+schema that will be exposed to the end user.  
+
+Applying fulltext search indexes across entities will be a powerful
```
Jannis: `Applying fulltext search indexes` -> `Enabling fulltext search`

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+The proposed process of adding one or more fulltext search index to an Entity 
+involves adding one or more fulltext directive to the `_Schema_` type in the 
+subgraph's GraphQL schema.  Each fulltext definition will have four required
+top level parameters: `name`, `language`, `algorithm`, and `include`. 
+The fulltext search definitions will be used to generate queries on the GraphQL
+schema that will be exposed to the end user.  
+
+Applying fulltext search indexes across entities will be a powerful 
+abstraction allowing users to search across all relevant entities in one
+query. An index search across entities will by definition have polymorphic
```
Jannis: `An index search across entities will...` -> `Such a search will...`

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+results, so in that case a union type will also be generated in the schema 
+for the query result items.
```
Jannis: > results, so in that case a union type will also be generated in the schema for the query result items.

->

> results. To address this, a union type will be generated in the API schema for fulltext search results.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+The fulltext search definitions will be used to generate queries on the GraphQL
+schema that will be exposed to the end user.  
+
+Applying fulltext search indexes across entities will be a powerful 
+abstraction allowing users to search across all relevant entities in one
+query. An index search across entities will by definition have polymorphic
+results, so in that case a union type will also be generated in the schema 
+for the query result items. 
+
+Verification of the index definition will ensure that all fields referenced
```
Jannis: `Verification` is no the right term. `Validation` is. There's also frequent use of the term `index` in this paragraph.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+Applying fulltext search indexes across entities will be a powerful 
+abstraction allowing users to search across all relevant entities in one
+query. An index search across entities will by definition have polymorphic
+results, so in that case a union type will also be generated in the schema 
+for the query result items. 
+
+Verification of the index definition will ensure that all fields referenced 
+are valid String type fields. To be clear, the proposed interface will add 
+the fulltext search index definitions to the subgraph deployment hash, so 
+adding or updating a fulltext search index will require an update to the subgraph.
```
Jannis: The `To be clear` part is obvious. It's part of the schema, the schema affects the subgraph ID. We can drop the entire sentence.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+  - _standard ranking_: ranking based on the number of matching lexemes.     
+  - _cover density ranking_: Cover density is similar to the standard 
+    fulltext search ranking except that the proximity of matching lexemes 
+    to each other is taken into consideration. This function requires 
+    lexeme positional information to perform its calculation, so it ignores 
+    any "stripped" lexemes in the index.            
+
+## Detailed Design
+
+### Subgraph Definition
```
Jannis: Definition -> Schema. `Subgraph Definition` is not a term we use (other than informally perhaps).

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+```graphql
+type _Schema_ 
+  @fulltext(
+    name: "media"
+    ...
+  )
+  @fulltext(
+    name: "search",
+    language: "english",
+    algorithm: "ranked",
```
Jannis: If we generate a `_FulltextAlgorithm` or similar enum type, we can use `ranked` instead of `"ranked"` here:
```
enum _FulltextAlgorithm {
  ranked
}
```

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+        ]
+      }
+    ]
+  )
+```
+
+The schema generated from the above definition: 
+```graphql
+union FulltextMediaResultItem = ...
+union FulltextSearchResultItem = Band | Album | Musician
```
Jannis: I think we should call these `_FulltextMediaEntity` and `_FulltextSearchEntity`.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+  media...
+  search(text: String!, first: Int, skip: Int, block: Int): [FulltextSearchResultItem!]!
+}
+```
+
+### GraphQL Query interface
+
+End users of the subgraph will have access to the queries generated for the
+fulltext search indexes alongside the other queries generated for each 
+entity in the subgraph.  In the case of a fulltext search index that spans
```
Jannis: Queries are not generated...

Jannis (changes requested):

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+multiple entities, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) 
+may be used in the query to deal with the polymorphic result items. In the 
+front-end the `__typename` in each result item can be used to render the 
+components accordingly.
```
Jannis: Which components? We could just say `the `__typename` field can be used to distinguish the concrete entity types of the returned results`.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+
+End users of the subgraph will have access to the queries generated for the
+fulltext search indexes alongside the other queries generated for each 
+entity in the subgraph.  In the case of a fulltext search index that spans
+multiple entities, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) 
+may be used in the query to deal with the polymorphic result items. In the 
+front-end the `__typename` in each result item can be used to render the 
+components accordingly.  
+
+In the "text" parameter supplied to the query there will be one special
```
Jannis: `"test"` -> ``` `text` ```

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+End users of the subgraph will have access to the queries generated for the
+fulltext search indexes alongside the other queries generated for each 
+entity in the subgraph.  In the case of a fulltext search index that spans
+multiple entities, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) 
+may be used in the query to deal with the polymorphic result items. In the 
+front-end the `__typename` in each result item can be used to render the 
+components accordingly.  
+
+In the "text" parameter supplied to the query there will be one special 
+operator supported, the proximity operator. Several operators are available:
```
Jannis: One special operator or several operators?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+entity in the subgraph.  In the case of a fulltext search index that spans
+multiple entities, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) 
+may be used in the query to deal with the polymorphic result items. In the 
+front-end the `__typename` in each result item can be used to render the 
+components accordingly.  
+
+In the "text" parameter supplied to the query there will be one special 
+operator supported, the proximity operator. Several operators are available: 
+and, or, and proximity (`&`, `|`, `<->`.) The proximity operator allows 
+one to specify max distance between search terms: 3 words apart → `<3>`.
```
Jannis: one -> clients
max distance -> the maximum distance

``` 3 words apart -> `<3>` ``` -> ``` `foo<3>bar` is equivalent to requesting that `foo` and `bar` are three words apart ideally`?

File: rfcs/0004-fulltext-search.md [hunk 37]
```diff
@@ -0,0 +1,275 @@
...
+  )
+```
+
+The schema generated from the above definition: 
+```graphql
+union FulltextMediaResultItem = ...
+union FulltextSearchResultItem = Band | Album | Musician
+type Query {
+  media...
+  search(text: String!, first: Int, skip: Int, block: Int): [FulltextSearchResultItem!]!
```
Jannis: `block` is not an `Int`. I think it's `Block_height`.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+fulltext search features built in to PostgreSQL.
+
+A FullText search field will get its own column in the Entity table just 
+like any other field; however, the data stored will be the result of the 
+lexical, morphological analysis of text documents performed on the fields 
+included in the index. The fulltext search field will be created using the 
+Postgres ts_vector function and will be indexed using a GIN index. The subgraph 
+developer will define a ranking algorithm to be used to sort query results,
+so the end-user facing API remains easy to use without any requirement to 
+understand the ranking algorithms.
```
Jannis: The last time we discussed this, we discussed using a separate table for this data. I'm ok either way, one extra column per entity per fulltext search field could work as well. Just wanted to double-check which one we're choosing.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+so the end-user facing API remains easy to use without any requirement to 
+understand the ranking algorithms.   
+
+Limitations of Postgres fulltext search indexes and search algorithms: 
+  - only certain languages available (expandable with plugins like [PGroonga](https://pgroonga.github.io/)), 
+    - **languages supported out of the box**: 
+        Danish, Dutch, English, Finnish, French, German, Hungarian, Italian, 
+        Norwegian, Portuguese, Romanian, Russian, Spanish, Swedish, and Turkish.
+  - select algorithms are available, and 
+  - limited to PostgreSQL storage.
```
Jannis: Do we care about Postgres here? Isn't it an implementation detail? Aren't we starting only with English initially?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+This proposal does not change any existing interfaces, so no migrations 
+will be necessary for existing subgraph deployments. 
+
+## Drawbacks and Risks
+
+There is little risk in implementing this feature.  The proposed solution 
+uses native Postgres fulltext features and there is a nonzero probability 
+this choice results in slower than optimal write and read times; 
+however the tradeoff in implementation time/complexity and the existence 
+of production use case testimonials tempers my apprehension here.
```
Jannis: Another risk is that the fulltext search ability will not be translatable to future phases of the network.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Drawbacks and Risks
+
+There is little risk in implementing this feature.  The proposed solution 
+uses native Postgres fulltext features and there is a nonzero probability 
+this choice results in slower than optimal write and read times; 
+however the tradeoff in implementation time/complexity and the existence 
+of production use case testimonials tempers my apprehension here. 
+
+Are there any desired "fulltext search like" capabilities that I've missed in this proposal?
```
Jannis: That would be something for the open questions section. :wink:

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+for indexes. This design moved the definition of custom indexes out of the 
+subgraph definition; the subgraph would be deployed without them and they 
+could be added later using a new Graph Explorer interface (in Hosted-Service context) 
+or a JSON-RPC request directly to a Graph Node. One of the benefits of this 
+design is that the definition of uniqueness for a subgraph does not include 
+the custom indexes, so a new subgraph deployment and corresponding syncing 
+work does not have to be added in order to create or update an index. 
+However, it also introduces significant added complexity: will a separate 
+query marketplace and discovery registry then be required for finding nodes 
+with your needed subgraph-index combination.
```
Jannis: Is this a question?

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+however the tradeoff in implementation time/complexity and the existence 
+of production use case testimonials tempers my apprehension here. 
+
+Are there any desired "fulltext search like" capabilities that I've missed in this proposal?   
+
+## Alternatives
+
+An alternative design for the feature would allow more flexibility for 
+Graph Node operators in their index implementation and create a marketplace 
+for indexes. This design moved the definition of custom indexes out of the
```
Jannis: I find the mix of future and past tense here confusing. Perhaps describe the actual design here briefly. Or make it clear that it _is_ described here, if that's the case.

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,275 @@
...
+and the query result.  Can we find a healthy compromise between being 
+prescriptive about the indexes and algorithms in order to allow formal 
+verification and allowing indexer node operators to experiment with 
+algorithms and indexes in order to continue to improve query speed and results? 
+
+Since a fulltext search field is purely derivative of other Entity data 
+the addition or update of an @index field does not require a full blockchain 
+resync, rather the index itself just needs to be rebuilt. 
+How can we allow fulltext search index (and eventually other index types) 
+updates without requiring a full subgraph resync?
```
Jannis: We can't currently. For the moment, I feel that's ok. Should we rename `@index field` to `@fulltext directive`?

rfcs/approved.md
```diff
@@ -2,3 +2,4 @@
 
 - [RFC-0001: Subgraph Composition](./0001-subgraph-composition.md)
 - [RFC-0002: Ethereum Tracing Cache](./0002-ethereum-tracing-cache.md)
+- [RFC-0004: Fulltext Search](./rfcs/0002-fulltext-search.md)
```
Jannis: 0002 -> 0004

Jannis (changes requested):

rfcs/0004-fulltext-search.md
fordN: Do we not create items (queries) on the Query type?

rfcs/0004-fulltext-search.md
fordN: I agree the Postgres implementation details should be in the plan not here.  

I remember starting with English only being mentioned on the PR discussion call, but I thought it was just an option in case we need to simplify. Is there any reason not to include the option for other languages if it doesn't require more implementation work?

rfcs/0004-fulltext-search.md
fordN: Isn't that an implementation detail?

Jannis (approved):

rfcs/0004-fulltext-search.md
```diff
@@ -0,0 +1,261 @@
...
+
+```graphql
+type _Schema_ 
+  @fulltext(
+    name: "media"
+    ...
+  )
+  @fulltext(
+    name: "search",
+    language: "english",
```
lutter: We should use [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) codes for the language (like `en` for English, `de` for German etc.) Everything else on the web uses them, too.

fordN: I'm open to that though I think the full names are a bit more clear because they are self-describing.

lutter: They are contrary to anywhere else where systems talk about human languages (e.g., HTTP, localization libraries, etc.) which all use the ISO codes. It's also not quite true that `French` is called `French` or `Spanish` `Spanish` ;)

fordN: One counter example is Postgres which doesn't use the ISO codes for its fulltext search language configuration values, but I like the idea of using a standard that isn't english centric.  
I'll update the RFC to use `ISO 639-1` codes for the language value.

PR #13, opened 2020-01-24
PLAN-0001: Rewrite to use a lateral join instead of window functions

lutter:


lutter:
Just realized there had been no activity on this. Could somebody review/approve?

Jannis (approved):

PR #17, opened 2020-03-17
PLAN-0004: Subgraph grafting

lutter:


engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,143 @@
...
+    # Option 2: (copy) The deployment starts with a copy of the data from
+    # the given subgraph at the given block
+    deployment:
+      mode: copy
+      subgraph: Qm...
+      block: 123456
+    # Option 3: (reuse) The deployment reuses the data of an existing subgraph
+    # in place. This mode implicitly deletes the source deployment, and any
+    # subgraph version using the source deployment will be switched to use the
+    # new deployment
```
Jannis: At least on the hosted service, this could mean messing with existing deployments that other people have deployed copies of under different subgraph names and may be relying on. I don't think we want that.

engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,143 @@
...
+
+- For `copy` and `reuse` it is a deploy-time error if the source subgraph
+  does not exist or has not indexed the given `block` yet
+- To start with, for grafting, the source subgraph and the grafted subgraph
+  must have identical GraphQL schemas. We can expand/relax this requirement
+  over time.
+- It is an error to graft onto a source subgraph that uses JSONB storage
+  (i.e., the source subgraph must use relational storage)
+- For expediency, we will first implement `copy` and see if `reuse` is
+  actually needed - the big advanatge of `reuse` is that it does not
```
Jannis: Typo: advanatge -> advantage

engineering-plans/0004-subgraph-grafting.md
Jannis: Having said that, we could allow it for subgraph deployments that are exclusively used by the subgraph name we're operating against. What do you think? (It may be worth adding a note about this.)

engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,143 @@
...
+  require copying data and will therefore be faster to initialize a
+  subgraph, but it remains to be seen how much of a factor that is in
+  practice
+
+## Implementation
```
Jannis, line 69: We'll need to add a bit of manifest validation in graph-cli and graph-node. Not much luckily.

engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,143 @@
...
+The plan is described above under 'Implementation'
+
+## Open Questions
+
+- Should subgraphs deployed with `copy` or `reuse` be marked somehow as it
+  will be very hard to recreate them in other `graph-node` installations?
+  Recreating such a subgraph would require deploying the source subgraph,
+  letting it index past the source block and then deploying the target
+  subgraph. Since the source subgraph could itself be the result of
+  grafting, it might be necessary to perform these operations recursively
```
Jannis: Marked as in "marked in the explorer"? Apart from that, wouldn't the grafted deployments be marked in the database via their deployment mode?

Jannis: Recreating a grafted subgraph on a different graph node is indeed tricky. I have no good ideas here yet...

Jannis (changes requested):

engineering-plans/0004-subgraph-grafting.md
lutter: The more I think about `reuse`, the more I think we should not do it since it has all kinds of downsides and complications, and the main advantage is to save time on initial setup. We should definitely not do it unless we run into a situation where `copy` is not fast enough, and even then we might not want to allow it in shared setups like hosted.

Jannis: I'm on board with that.

That3Percent: I don't think that having modes and deploy time errors is a good way to surface the API in the first place. Semantically, "deploy mode" has no impact on the result of the expected results of a query for a particular block. Re-use, copy, and the like are internal implementation details of graph-node itself, and what decisions are made here may be resulting from external state beyond the subgraph (like whether or not an indexer has both the old and new version of the subgraph deployed). This whole section should be changed to reflect how grafting works semantically and not operationally.

engineering-plans/0004-subgraph-grafting.md
leoyvens: It does defeat the entire point if the node doesn't have the base subgraph synced to the required block. In those cases the deployment should just be rejected imo.

engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,143 @@
...
+    # new deployment
+    deployment:
+      mode: reuse
+      subgraph: Qm...
+      block: 123456
+
+- For `copy` and `reuse` it is a deploy-time error if the source subgraph
+  does not exist or has not indexed the given `block` yet
+- To start with, for grafting, the source subgraph and the grafted subgraph
+  must have identical GraphQL schemas. We can expand/relax this requirement
```
That3Percent: For this to be useful we may need to support a limited amount of schema changes at the outset - like converting a non-nullable field to a nullable field.

engineering-plans/0004-subgraph-grafting.md
That3Percent: I disagree that this is hard to recreate - subgraphs grafted upon should be on IPFS. That's what deploying means in The Network.

I also disagree that it defeats the entire point if the subgraph is not synced on a particular deployment. It may be synced on some nodes and not others, or synced locally and not in production. Quick syncing is not the only use for grafting, and quick syncing may only take effect in some environments but that doesn't mean it should fail in others where there is less benefit.

No, subgraphs deployed with `copy` or `reuse` should not be "marked somehow".

That3Percent (changes requested):
Comments below

engineering-plans/0004-subgraph-grafting.md
lutter: I view grafting as more than just an implementation detail, especially `reuse` as it leads to visibly different behavior (one subgraph disappears, another one appears)

But even with `copy`, deploy mode influences what data you ultimately put in your subgraph, e.g., because grafting onto a different block will in general lead to different data.

Having said that, how would you rewrite this to describe it at a more semantic level?

engineering-plans/0004-subgraph-grafting.md
lutter: Good point

engineering-plans/0004-subgraph-grafting.md
lutter: I wasn't going to store deployment mode explicitly, just extract it from the manifest on creation and act accordingly. We could store it in the database, either in the `SubgraphDeployment` or the `SubgraphManifest`, though it's not clear to me what we need to store it for.

lutter: What you are suggesting would require quite a bit more machinery: when a grafted subgraph is deployed, we would first need to (recursively) deploy its source, let that sync to the desired block (what if it fails before then?), and then trigger the deploy of the graft. We'd then also need to handle reverts across the grafting block - I should clarify this, but the behavior I plan on is that the revert would just happen within the data and logic of the grafted subgraph.

The main use case for this feature is to fix subgraph errors quickly, not as a general feature for code management. When it's planned, a subgraph writer can achieve the same effect as grafting with simple logic in their mappings, which is also much more flexible as the change can be triggered on more than just a block number. Grafting is really for situations where you have already synced for a long period of time, discover a bug in your mapping, and can address that by resyncing only a small portion of the subgraph rather than starting over.

leoyvens:
We should make sure validate that the block number is past the reorg threshold on deploy.

engineering-plans/0004-subgraph-grafting.md
That3Percent: In my mind a graft is a completely new subgraph, really exactly like a fork. It semantically executes one function up to a point, then another function from then on. Whether or not the results of that execution are cached and how is an implementation detail. (Eg: An index node that re-indexed from scratch for each query while wildly inefficient would not be semantically incorrect). Where you get the data from the previous blocks functions executing is a question of how the cache is implemented.

The concept of one subgraph disappearing and another appearing has no place in that mental model.

If anything, the graft is more like a data source than anything else.
Eg:
```toml
dataSources:
    - kind: fork
       subgraph: '0x0'
    - kind: ethereum/contract
       startBlock: 0 <- the block that the new function starts executing on
```
(The particulars of where this information go in subgraph.yaml or how this concept is open to debate)

The extent of the semantics that need to be described are "execute one function up to here, then a different function from then on" is easier to understand than what is written, and more powerful. The old subgraph still exists, and may or may not still be queried depending on whether it is deployed.

This model may enable a kind of backward compatible rollout of subgraph updates, or updates to live subgraphs that occur from a future block, as well as the given use case of rapidly deploying fixes.

I feel though I'm on a very different page then the rest of the team here. In the name of forward progress, I'm willing to just give up pressing this issue if that would be more useful.

That3Percent (approved):

engineering-plans/0004-subgraph-grafting.md
lutter: I think the disconnect is in viewing the subgraph data as precious in some way: other than for the fact that reindexing takes time, that data is not precious in the same way that data in other applications is - all subgraph data is derivative and not original, and can be reconstructed. Grafting is an attempt at making the time it takes to reindex/address a problem shorter.

I don't think it is a good tool to express the evolution of logic of subgraph mappings, since it's much harder to see how behavior evolved with grafts than if that was expressed explicitly in the mappings; i.e., with grafting, you have to compare your subgraph code at commits X and Y (or at different tags or branches) to see how logic has changed whereas if that's done in the mapping code you'd be looking at an if statement somewhere.

lutter:
> We should make sure validate that the block number is past the reorg threshold on deploy.

I feel that that would slow things down too much, as that would currently mean you have to wait 25 minutes before you can deploy your graft. I feel in this case, we should probably just let it slide as the time it will take to notice that the subgraph is broken, come up with a fix and deploying that should take enough time to make reverting across the graft the exception.

lutter:
Added detailed steps under `Implementation Plan`

leoyvens:
@lutter It's unclear to me what the plan is for reverts. Even if they are the exception, they should still be deterministic.

lutter:
> @lutter It's unclear to me what the plan is for reverts. Even if they are the exception, they should still be deterministic.

I agree. I just added a statement that we will only graft onto final blocks, i.e. that the chain head must be at least `REOGR_THRESHOLD` blocks beyond the graft point, and that the graft point must be on the main chain.

engineering-plans/0004-subgraph-grafting.md
```diff
@@ -0,0 +1,170 @@
...
+  actually needed; `reuse` is very invasive and can interact badly with
+  other features (e.g., sharing of subgraph deployments under several
+  names). The big advantage of `reuse` is that it does not require copying
+  data and will therefore be faster to initialize a subgraph, but it
+  remains to be seen how much of a factor that is in practice
+- For `copy` and `reuse` it is a deploy-time error if the source subgraph
+  does not exist.
+- For `copy` and `reuse`, the block up to which we use the source subgraph
+  must be final. It must therefore be at least `ETHEREUM_REORG_THRESHOLD`
+  blocks behind the chain head, and be on the main chain.
```
Jannis: This is a problem, because it implies a minimum downtime of `ETHEREUM_REORG_THRESHOLD` blocks.  On the hosted service (current reorg threshold: 200 blocks), this corresponds to roughly 50 minutes (at 15s block times).

Can't we require a block hash to continue from, identify what block number that corresponds to, fail if we can't, and otherwise copy the source subgraph based on that number? If this block were to be reorged later, the subgraph would handle that just like the original.

Jannis (changes requested):

engineering-plans/0004-subgraph-grafting.md
Jannis: After talking to @leoyvens, I realize that the situation is more complicated.

When grafting against block `N` but block `N` gets reorg'ed later, your grafting base block doesn't even exist anymore. Which is weird. Suddenly you have a different base block...?

In a sense, grafting demands that the base block is treated as final. Since we can't guarantee that, @leoyvens suggested failing whenever the grafting base block is reorg'ed, which I agree with.

This would allow us to use more recent blocks for grafting. The likelihood that a block mined 5 minutes ago is already very, very low, so in reality, we should very rarely see a grafting block being reorganized and causing such a failure.

lutter: I agree that that is a pragmatic compromise between making people wait forever to be allowed to graft and risking undefined behavior. And that also gives a reason for storing the base block for a graft with the subgraph, as that needs to be checked now for each subgraph for each revert.

As an aside, for grafting, it's enough to deal with block numbers rather than hashes as they are unique in the context of any given subgraph.

lutter:
Updated with the latest discussion; the two additional tasks add ~ 1 day to the estimates

Jannis (approved):
LGTM

PR #18, opened 2020-03-19
RFC: Error handling and indexing status API improvements

leoyvens:
[RFC](https://github.com/graphprotocol/rfcs/blob/leo/error-handling/rfcs/0000-error-handling.md)

### DIFF
--- a/rfcs/0000-error-handling.md
+++ b/rfcs/0000-error-handling.md
@@ -1,0 +1,144 @@
+# RFC-0000: Subgraph error handling and reporting
+
+<dl>
+  <dt>Author</dt>
+  <dd>Leonardo Y. Schwarzstein</dd>
+    <dt>RFC pull request</dt>
+  <dd>URL</dd>
+
+  <dt>Date of submission</dt>
+  <dd>2020-03-19</dd>
+
+  <dt>Date of approval</dt>
+  <dd>YYYY-MM-DD</dd>
+
+  <dt>Approved by</dt>
+  <dd>First Person, Second Person</dd>
+</dl>
+
+## Contents
+
+<!-- toc -->
+
+## Summary
+
+Synced subgraphs no longer stop syncing on errors. Failed subgraphs return an error along with the data when queried.
+
+The indexing status API is improved and documented.
+
+## Goals & Motivation
+
+Currently, an error in a WASM trigger handler will cause a subgraph to be marked as failed and stop syncing. The proposal is for it still be marked as failed, but discard any changes made by the failed handler and attempt to continue. In the cases where the failure is catastrophic and all handlers start failing, this will not do much, but in the cases where the subgraph can continue operational with a bit of missing data, this will allow applications to continue running on up-to-date data and give developers more time to fix their subgraph.
+
+Less catastrophic failures do have a downside: It's harder to notice that anything is wrong at all. We'll include an error in the graphql response for failed subgraphs. We will improve and document the indexing status API, so that developers can programmatically check details on a subgraph's health and build their own alerts and dashboards.
+
+## Urgency
+
+Sooner is better than later.
+
+## Terminology
+
+No new terminology, but the semantics of the "Failed" status are altered.
+
+## Detailed Design
+
+### Error handling
+
+If a WASM handler terminates with a trap, that is an error. The node discards all entity operations and data source operations done by the handler and continue syncing.
+
+For subgraphs that are not yet synced, any handler error halts syncing, so errors in pending versions will not make their way to the current version. 
+
+When the node starts, we currently reset the failed flag on subgraphs. The flag is no longer be reset on synced subgraphs, since the subgraph will already be past the errors.
+
+Any graphql queries to a failed subgraph will include an error of the form:
+
+```
+{
+	message: "Subgraph failed"
+}
+```
+
+So that applications will always be aware that queries may have inconsistent data. Developers should refer to the indexing status API for details on the errors.
+
+### Indexing Status API
+
+The indexing status API will be revamped.
+
+The subgraph indexing status API currently has the top level query:
+
+    indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
+
+This is changed to:
+
+    indexingStatusesForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+
+Which returns null if the name is not found, and returns the status for the current version of the subgraph otherwise.
+
+A field `failedAfter`  is added to record the block previous to the first error.
+
+Errors are made more than just a string with the `Error` type.
+
+The complete indexing status API schema is:
+
+```graphql
+type Query {
+  indexingStatusesForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+  indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+}
+
+type SubgraphIndexingStatus {
+  subgraph: String!
+  synced: Boolean!
+  errors: [Error!]! # Sorted from first to last
+  chains: [ChainIndexingStatus!]!
+  node: String!
+}
+
+# Fields are moved from `EthereumIndexingStatus` to the interface.
+interface ChainIndexingStatus {
+  network: String!
+  chainHeadBlock: Block
+  earliestBlock: Block
+  latestBlock: Block
+  failedAfter: Block
+}
+
+type EthereumIndexingStatus implements ChainIndexingStatus { }
+
+# Renamed from `EthereumBlock` to `Block`
+type Block {
+  hash: Bytes!
+  number: BigInt!
+}
+
+type Error {
+	message: String!
+	
+	# Context for the error.
+	network: String!
+	block: Block!
+	transaction: Bytes
+	handler: String
+}
+```
+
+Since the indexing API is a view to the subgraph metadata, an `Error` type is also added to the metadata where subgraph errors are recorded.
+
+## Compatibility
+
+This does incompatible changes to the indexing status API, however that API was so far experimental and undocumented.
+
+The behavior of synced, failed subgraphs is changed to keep indexing. Our consensus is that the current behavior is useful to nobody and this can only be an improvement.
+
+## Drawbacks and Risks
+
+- Changing the behavior of failed subgraphs has risks. There may be unanticipated consequences which may slip through testing. We may have gotten our rationale wrong and users may find the new behavior worse for factors we did not consider.
+
+## Alternatives
+
+- Don't change the default behavior, and put make the "continue on failure" mode opt-in through a flag on the manifest. With the possible intention of making it the default in the future. This exposes us to less risk when rolling this out. The downsides are that adoption and feedback will be slower and developers will need to update their subgraphs in order to benefit.
+
+## Open Questions
+
+- The field name `failedAfter` is up for bikeshedding, aternatives: `lastHealthyBlock`, `blockBeforeFailure`, `lastGoodBlock`...
+
--- a/rfcs/0000-error-handling.md
+++ b/rfcs/0000-error-handling.md
@@ -122,11 +122,11 @@ type Error {
 
 enum Health {
   "Subgraph syncing normally"
-  Healthy,
+  healthy,
   "Subgraph syncing but with errors"
-  Unhealthy,
+  unhealthy,
   "Subgraph halted due to errors"
-  Failed,
+  failed,
 }
 ```
 
@@ -149,6 +149,5 @@ The behavior of synced subgraphs when encountering an error is changed to keep i
 
 ## Open Questions
 
-- The field name `lastHealthyBlock` is up for bikeshedding, aternatives: `failedAfter`, `blockBeforeFailure`, `lastGoodBlock`...
-- Do we want to keep the `node` field in the indexing status API? It feels too iternal.
+No unresolved questions.
 
--- a/rfcs/0000-error-handling.md
+++ b/rfcs/0000-error-handling.md
@@ -22,15 +22,15 @@
 
 ## Summary
 
-Synced subgraphs no longer stop syncing on errors. Failed subgraphs return an error along with the data when queried.
+Synced subgraphs no longer halt on mapping errors, but are marked as unhealthy. Unhealthy subgraphs return an error along with the data when queried.
 
 The indexing status API is improved and documented.
 
 ## Goals & Motivation
 
-Currently, an error in a WASM trigger handler will cause a subgraph to be marked as failed and stop syncing. The proposal is for it still be marked as failed, but discard any changes made by the failed handler and attempt to continue. In the cases where the failure is catastrophic and all handlers start failing, this will not do much, but in the cases where the subgraph can continue operational with a bit of missing data, this will allow applications to continue running on up-to-date data and give developers more time to fix their subgraph.
+Currently, an error in a WASM trigger handler will cause a subgraph to be marked as failed and stop syncing. The proposal is for it to discard any changes made by the failed handler, mark the subgraph with the new 'unhealthy' status and continue syncing. In the cases where the error is catastrophic and all handlers start erroing, this will not do much, but in the cases where the subgraph can continue operational with a bit of missing data, this will allow applications to continue running on up-to-date data and give developers more time to fix their subgraph.
 
-Less catastrophic failures do have a downside: It's harder to notice that anything is wrong at all. We'll include an error in the graphql response for failed subgraphs. We will improve and document the indexing status API, so that developers can programmatically check details on a subgraph's health and build their own alerts and dashboards.
+Less catastrophic errors do have a downside: It's harder to notice that anything is wrong at all. We'll include an error in the graphql response for unhealthy and failed subgraphs. We will improve and document the indexing status API, so that developers can programmatically check details on a subgraph's health and build their own alerts and dashboards.
 
 ## Urgency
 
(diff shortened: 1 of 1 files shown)


rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+
+The indexing status API will be revamped.
+
+The subgraph indexing status API currently has the top level query:
+
+    indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
+
+This is changed to:
+
+    indexingStatusesForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
```
Jannis: I'd rather add a new, separate resolver for just the current version:
```
indexingStatusForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
```

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+```graphql
+type Query {
+  indexingStatusesForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+  indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+}
+
+type SubgraphIndexingStatus {
+  subgraph: String!
+  synced: Boolean!
+  errors: [Error!]! # Sorted from first to last
```
Jannis: This removes the `failed` flag because errors implicitly mean the subgraph is failing?

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+  number: BigInt!
+}
+
+type Error {
+	message: String!
+	
+	# Context for the error.
+	network: String!
+	block: Block!
+	transaction: Bytes
```
Jannis: Our error messages should already include the transaction. I'd prefer to code as little structural information about the underlying network in these errors as possible to keep us flexible. (Even if transactions will probably be a thing everywhere.)

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Changing the behavior of failed subgraphs has risks. There may be unanticipated consequences which may slip through testing. We may have gotten our rationale wrong and users may find the new behavior worse for factors we did not consider.
+
+## Alternatives
+
+- Don't change the default behavior, and put make the "continue on failure" mode opt-in through a flag on the manifest. With the possible intention of making it the default in the future. This exposes us to less risk when rolling this out. The downsides are that adoption and feedback will be slower and developers will need to update their subgraphs in order to benefit.
+
+## Open Questions
+
+- The field name `failedAfter` is up for bikeshedding, aternatives: `lastHealthyBlock`, `blockBeforeFailure`, `lastGoodBlock`...
```
Jannis: Heh, yeah, I was going to suggest `lastHealthyBlock`. 😄

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+Since the indexing API is a view to the subgraph metadata, an `Error` type is also added to the metadata where subgraph errors are recorded.
+
+## Compatibility
+
+This does incompatible changes to the indexing status API, however that API was so far experimental and undocumented.
+
+The behavior of synced, failed subgraphs is changed to keep indexing. Our consensus is that the current behavior is useful to nobody and this can only be an improvement.
```
Jannis: I wonder if `failed` is still the right term given the new behavior. If subgraphs keep going, `failing` or `unhealthy` could be better, more clear terms. What do you think?

Jannis (changes requested):

rfcs/0000-error-handling.md
leoyvens: Removed.

rfcs/0000-error-handling.md
leoyvens: I've added a `health: Health` field here and the `Health` enum.

rfcs/0000-error-handling.md
leoyvens: That's better, the name was still plural which made no sense.

rfcs/0000-error-handling.md
leoyvens: Switched to that.

rfcs/0000-error-handling.md
leoyvens: I've done a major change to the RFC to explicitly use a third "unhealthy" state for this.

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,154 @@
...
+	handler: String
+}
+
+enum Health {
+  "Subgraph syncing normally"
+  Healthy,
+  "Subgraph syncing but with errors"
+  Unhealthy,
+  "Subgraph halted due to errors"
+  Failed,
```
Jannis: Nit: Let's use lower-case enum values. It feels more consistent with all other enums we use in our GraphQL APIs (e.g. `orderDirection: asc`, `where: { something_gt: ... }`).

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Alternatives & Extensions
+
+- Don't change the default behavior, and put make the "continue on failure" mode opt-in through a flag on the manifest. With the possible intention of making it the default in the future. This exposes us to less risk when rolling this out. The downsides are that adoption and feedback will be slower and developers will need to update their subgraphs in order to benefit.
+- Future extension: Allow mappings to define a trap handler that catches traps on a specific trigger handler. The mapping can then decide whether the subgraph should fail, be set as unhealthy, or just continue normally.
+
+## Open Questions
+
+- The field name `lastHealthyBlock` is up for bikeshedding, aternatives: `failedAfter`, `blockBeforeFailure`, `lastGoodBlock`...
+- Do we want to keep the `node` field in the indexing status API? It feels too iternal.
```
Jannis: We added it for the indexer agent to be able to balance work across multiple index nodes. I think it's worth keeping.

Jannis (changes requested):
One tiny remark — I like the change introducing the `Health` enum!

rfcs/0000-error-handling.md
leoyvens: Done.

rfcs/0000-error-handling.md
leoyvens: Alright lets keep it.

Jannis (approved):

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,153 @@
+# RFC-0000: Subgraph error handling and reporting
+
+<dl>
+  <dt>Author</dt>
+  <dd>Leonardo Y. Schwarzstein</dd>
+    <dt>RFC pull request</dt>
+  <dd>URL</dd>
```
Jannis, line 7: Don't forget to add https://github.com/graphprotocol/rfcs/pull/18/files in place of `URL`.

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,153 @@
...
+  indexingStatusForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+  indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
+  indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+}
+
+type SubgraphIndexingStatus {
+  subgraph: String!
+  synced: Boolean!
+  health: Health!
+  errors: [Error!]! # Sorted from first to last
```
Jannis: Quick comment about the errors: there may be a _lot_ of them. A few ideas:

1. Only keep the most recent N errors around, for a fixed N (e.g. 100). I think this is bad, because you'll usually want to know where things started to go wrong and how. The first 100 would be similarly limiting.

2. Implement pagination a la `first`/`skip`. The main problem with this is that every new block may add new errors, so while you're paging through the errors, new ones will be added and the pages may jump around confusingly. What could work is starting `first` at the first/oldest error. Then you'd get stable pages and you'll get the most useful errors first, without needing to paginate.

Thoughts?

Jannis (changes requested):
I forgot about error pagination.

rfcs/0000-error-handling.md
leoyvens: I hadn't considered this problem. Paginating from the oldest error sounds good. This might not need a change in the RFC but it's something to keep in mind for the implementation.

Jannis: A third option would be to move the errors into separate `indexingErrorsForSubgraph` resolvers to which you have to pass the blocks for each network involved in the subgraph(currently only the Ethereum block hash) and mix that with pagination from the most recent error. Tying it to a specific block would make pagination stable, but starting from the most recent error is less useful. Approach 2 has my vote, I think.

Jannis: @leoyvens Yeah, agreed.

Jannis (approved):

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,153 @@
...
+
+Errors are made more than just a string with the `Error` type.
+
+The complete indexing status API schema is:
+
+```graphql
+type Query {
+  indexingStatusForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+  indexingStatusesForSubgraphName(subgraphName: String!): [SubgraphIndexingStatus!]!
+  indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
```
lutter, line 87: Can't we just have one `indexingStatuses` collection which takes a `subgraphs` and a `names` parameter (both `[String!]`) and returns an indexing status for each entry in those two parameters? Logically, there's one collection of statuses that we filter with different criteria (which in this case are implicitly or'd together)

rfcs/0000-error-handling.md
lutter: What's an upper bound on the number of errors we might encounter? Number of handlers * number of blocks? I assume dynamic data sources play into this, too, in which case each block could generate a ton of errors (thousands) We should predefine an error budget and if a subgraph goes beyond that still fail it. I don't think it helps anybody if we have subgraphs that can generate and store huge numbers of errors at every block - they'll most likely be useless and just eat up resources.

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+  indexingStatusesForSubgraphName(subgraphName: String!): SubgraphIndexingStatus
+  indexingStatuses(subgraphs: [String!]): [SubgraphIndexingStatus!]!
+}
+
+type SubgraphIndexingStatus {
+  subgraph: String!
+  synced: Boolean!
+  errors: [Error!]! # Sorted from first to last
+  chains: [ChainIndexingStatus!]!
+  node: String!
```
lutter, line 98: Is that information useful for users of the API? I am a little hesitant since it exposes internal information about the indexing infrastructure.

rfcs/0000-error-handling.md
```diff
@@ -0,0 +1,144 @@
...
+
+For subgraphs that are not yet synced, any handler error halts syncing, so errors in pending versions will not make their way to the current version. 
+
+When the node starts, we currently reset the failed flag on subgraphs. The flag is no longer be reset on synced subgraphs, since the subgraph will already be past the errors.
+
+Any graphql queries to a failed subgraph will include an error of the form:
+
+```
+{
+	message: "Subgraph failed"
```
lutter: Can we just use the `Health` status here? I'd prefer a machine-readable enum over essentially freeform text.

lutter (changes requested):

rfcs/0000-error-handling.md
leoyvens: We can limit the amount of errors stored, though I'd imagine that a successful handler will likely generate more data than an erroed one. I'm not in favor of having an error budget after which the subgraph is halted, thousands of operations could fail and there could still be useful ones going through.

rfcs/0000-error-handling.md
leoyvens: @Jannis manifested for keeping it in a thread above.

rfcs/0000-error-handling.md
leoyvens: As in "Subgraph healthy/unhealthy/failed" or just "healthy/unhealthy/failed"?

rfcs/0000-error-handling.md
leoyvens: Maybe but that doesn't seem like the easiest API to understand and use, I think `indexingStatusForSubgraphName` is exactly what most users want.

rfcs/0000-error-handling.md
lutter: Just "healthy/unhealthy/failed" - something that is guaranteed to stay that way, and whose exact text is part of the API so that people can reliably check for it.

rfcs/0000-error-handling.md
lutter: I am concerned with what that will do to the hosted service: right now, subgraphs that are clearly useless and fail do not cause any work after starting and failing them again. With this change, any subgraph, no matter how useless it is will continue to cause work. Currently, we have 686 assigned deployments that are working, and 112 that are failed. We would therefore increase the load in hosted by 16% for no real purpose whatsoever - if the failed subgraphs were important, they would have gotten fixed.

I think we need a better mechanism to avoid spending significant amounts of work on useless outcomes. Besides limiting the number of errors, I am not sure what else we can do to avoid that.

leoyvens: Yes, keeping subgraphs with errors running will require resources. To avoid that jump, we can make currently failed subgrahps stay failed when doing the migration.

We can establish some criteira by which we stop indexing an unhealthy subgraph on the hosted service, and possibly make that criteira built-in to the graph node.

For example, if every single handler errors for a configured number of blocks, then we stop indexing the subgraph. But because we don't yet have data on how many subgraphs would actually be stopped on a given criteria, I feel it's premature to establish one. If we run this on the hosted service for a while, we'll be better informed and able to experiment with criteira for failing unhealthy subgraphs.

rfcs/0000-error-handling.md
leoyvens: Should we use all caps `"UNHEALTHY"` so it looks more like a constant?

rfcs/0000-error-handling.md
Jannis: 1. I think it's fair to limit the number of errors we store to, say, 1000. And drop pagination even. As long as we also log the caught errors so we can establish a full list from the logs if needed.

2. The extra resources to keep failed subgraphs running are a good reason to make the feature opt-in, and keep the current failing behavior as the default. There are a few things we can do to prevent _everyone_ from using it, and thereby keep the resource use at bay, like stressing that this should only be used for production subgraphs. Warning about it in graph-cli. Things like that.

rfcs/0000-error-handling.md
Jannis: Right now it is. In the network `indexingStatuses(subgraphs: ["Qm..."])` becomes the more important API.

I agree with keeping names / IDs separate. One thing we may want to add to the `ForSubgraphName` resolvers are `current: Boolean!` and `pending: Boolean!` flags to allow clients to identify the current/pending versions. Right now, I think the order of statuses returned is supposed to communicate which version is current, but I'm not sure it's actually doing that...

rfcs/0000-error-handling.md
Jannis: I'd say lowercase is fine.

rfcs/0000-error-handling.md
leoyvens: Limiting to the first 1000 errors seems reasonable, I'll note in the RFC that the errors are limited.

It's true that not every subgraph needs this, and disabling it by default might save some hosted service resources.

The tradeoff being that some subgraph developers will find out the hard way that they need this, by going through an outage before finding out in our support channels that their issues can be mitigated by enabling this.

rfcs/0000-error-handling.md
leoyvens: Makes sense to have an API for the pending version, but I didn't understand exactly the API you suggest, we'd have two extra parameters, a `current` and a `pending` boolean? It seems you'd only need one. I think I prefer adding a new query:
```graphql
indexingStatusForSubgraphNamePending(subgraphName: String!): SubgraphIndexingStatus
```

Jannis: Ah, my suggestion was to add a `currentVersion` and `pendingVersion` flag to the returned indexing statuses. In the network the current version is determined by what's in the contracts. And pending versions are no longer a thing.

But I think it would be cleaner to add `indexingStatusForCurrentVersion` and `indexingStatusForPendingVersion` for subgraph names now and retire them in the network, like you suggest, and keep the indexing statuses "clear" of non-network features.

rfcs/0000-error-handling.md
Jannis: As long as we document that

* it is only recommended for production subgraphs
* but it _is_ recommended for production subgraphs

clear enough, I think that's ok. It could be good to have a "best practices for production subgraphs" section in the docs.

leoyvens: Alright, then we need to decide how this is specified in the manifest. For simplicity this can be a top-level flag on the manifest, applying to all data sources, though we could make it per-data source if we want to. Some brainstorming:

- Have a general `production` boolean, which we could possibly use for other behaviour that differs between a debug mode and a production mode.

- Have a more specific flag that reflects this particular behaviour. I'm not sure how to describe it, maybe `failOnMappingError: false`.

rfcs/0000-error-handling.md
leoyvens: I've updated this.

Jannis (approved):

PR #20, opened 2020-06-10
PLAN-0001: Add single child per parent as an important optimization

lutter:


engineering-plans/0001-graphql-query-prefetching.md
```diff
@@ -196,24 +209,30 @@ select c.*, p.id as parent_id
...
+##### Single child per parent
+
 ```sql
-expand_parents  := unnest({parent_ids}) as p(id)
-linked_children := p.id = c.{parent_field}
+select c.*, c.{parent_field} as parent_id
+  from children c
+ where c.{parent_field} = any({parent_ids})
+   and .. other conditions on c ..
+ limit {parent_ids.len} + 1
```
leoyvens, line 228: I couldn't find any big cases like this in the wild, the only one I found dealt with a very small amount of data. So I guess we don't need to focus too much on it at this point, but imo it would be more consistent with what we do in the other single child optimizations to join with the unnested ids, something like:
```
select c.*, c.{parent_field} as parent_id
  from unnest({parent_ids}) as p(id), children c
 where c.{parent_field} = p.id
   and .. other conditions on c ..
 limit {parent_ids.len} + 1
```

leoyvens (approved):
Just one comment on the queries. 

Nice that we're keeping this as a living document! Looking for a reference to here from the code I found this on the graph-node source:
```
/// Details of how query generation for `EntityQuery` works can be found
/// in `docs/implementation/query-prefetching.md`
```
We should update that comment to point to this engineering plan instead. Or maybe better, move this plan into the graph-node repo.

PR #21, opened 2021-02-26
Multi-blockchain support

leoyvens:
This builds on #8. The 'Subgraph Indexing' section has been rewritten and gained a [companion diagram](https://whimsical.com/multi-blockchain-13ZinTmYAUn5YknGjcZc4e).


rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+
+The feature proposed in this RFC is referred to as **multi-blockchain support**.
+Other terms that are introduced in this RFC or are relevant when talking about
+it are listed below.
+
+- **Network** - a decentralized technology or similar, such as Ethereum,
+  Substrate, or IPFS.
+- **Blockchain** - a blockchain technology, such as Ethereum or Substrate.
+- **Chain** - a specific instance of a _blockchain_, such as Ethereum mainnet.
+- **Chain subgraph** - a built-in subgraph that indexes data intrinsic to a
```
yanivtal, line 61: Should we make this more generic and call it a Network Subgraph?

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+RFC, only Ethereum and blockchains compatible with the Ethereum JSON-RPC API are
+supported.
+
+This feature unlocks the same use cases on other blockchains that The Graph has
+made possible on Ethereum, including indexing historical data, interacting with
+smart contracts (if supported by the respective blockchain), handling reorgs (if
+applicable) and triggering off of blockchain-specific features like events or
+transactions.
+
+## Urgency
```
yanivtal, line 45: This Urgency section looks outdated. Could revise a bit to say that other chains are gaining traction this year and we want to help developers take advantage of the unique properties of each chain.

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+
+This section specifies how all of these components are updated to support other
+chains and what the integration of additional chains looks like. With this
+feature implemented, subgraph developers will be able to define new types of
+data sources for supported chains. Indexers will be able to configure which of
+these chains they want to enable.
+
+### Subgraph Manifest
+
+The structure of the subgraph manifest is already desigend for being extended
```
yanivtal, line 85: desigend  -> designed

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+  ...
+```
+
+Manifests can be migrated from the old structure to this automatically. It will
+however require code changes and a database migration to support top-level ABIs.
+
+### Code Generation
+
+Code generation for schemas remains untouched by multi-blockchain support. Code
+generation for ABIs is extended to support more ABIs than just Ethereum's. Based
```
yanivtal, line 142: Should keep in mind that some blockchains like Near don't have ABIs

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+// - substrate.Block
+// - ...
+```
+
+Most, but not all of these, are internal to code generation. The same goes for
+low-level types such as Ethereum's `uint256` or `bytes32` type. These are
+already sufficiently abstracted behind `Bytes`, `BigInt`, `BigDecimal` and the
+code generated for Ethereum tuples.
+
+Additional types may be necessary for other chains, but it expected that the
```
yanivtal, line 181: it -> it's

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+Data enters the system by being requested from the chain
+itself, for example through JSON-RPC calls. The __adapters__ are the components responsible for
+interacting directly with the chain. They are broken up into specialized traits for each component
+that requires an adapter, such as the `IngestorAdapter` and `TriggersAdapter`. For performance, the
+adapter may cache and index into the local database the data pulled from the chain. Each chain will
+have its own DB schema (aka namespace) under which adapters may create any necessary custom DB
+tables.
+
+The __block ingestor__ and the __chain store__ will be reused across chains. There will be one
+instance of each per chain, an `IngestorAdapter` will provide access chain specific data. Each
```
yanivtal, line 230: access chain -> access to chain

yanivtal (commented):
Looking great. Just some minor comments

rfcs/0005-multi-blockchain-support.md
lutter: We are trying to use the vocabulary that `Network` is a certain blockchain technology, and `Chain` the concrete incarnation of that technology. Calling it a 'Network Subgraph' would muddy this distinction, since the subgraph is specific to a chain. IOW, with what we have, Ethereum is the network, and `kovan` a chain, and the subgraph would be for `kovan` data. Though all this is moot for the time being since such a subgraph is an optional addition that we'll work on after we have the basics of multiblockchain right.

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+trait HostFns<C: Blockchain> { /* ... */ }
+
+trait BlockStream<C: Blockchain> {
+      fn new(&self, start_block: BlockPointer, data_sources: Vec<C::DataSource>, trigger_adapter: Arc<C::TriggerAdapter>) -> Result<Self, Error>;
+}
+
+/// Represents a blockchain supported by the node.
+trait Chain {
+    /// The DB schema for the `blocks` table and any custom tables.
+    const DB_SCHEMA: &'static str;
```
lutter, line 305: In the implementation, I would like to keep this more flexible, and probably have a callback `initialize_chain_store_schema` or some such that will be called whenever a chain needs to set up its database schema.

rfcs/0005-multi-blockchain-support.md
```diff
@@ -0,0 +1,487 @@
...
+
+Data enters the system by being requested from the chain
+itself, for example through JSON-RPC calls. The __adapters__ are the components responsible for
+interacting directly with the chain. They are broken up into specialized traits for each component
+that requires an adapter, such as the `IngestorAdapter` and `TriggersAdapter`. For performance, the
+adapter may cache and index into the local database the data pulled from the chain. Each chain will
+have its own DB schema (aka namespace) under which adapters may create any necessary custom DB
+tables.
+
+The __block ingestor__ and the __chain store__ will be reused across chains. There will be one
```
lutter, line 229: There's an inherent chain-specific dependency between the adapter and the block stream, which communicate with each other through data in the database: the adapter writes data in a chain-specific format into the chain store, and the block stream reads it. A lot of that can probably be handled inside the implementation of the chain itself, but the current `ChainStore` trait will at least need to be parametrized on the type of the block (currently `EthereumBlock`) Ideally though, we could remove those methods from the public chain interfaces and keep the notion of what a block looks like within the chain implementation.

lutter (approved):
Nice! There's some details to work out around the chain store, but I think that can happen as we refactor the current ethereum support.

rfcs/0005-multi-blockchain-support.md
leoyvens: Good reminder but we don't need to work out the details right now so I'll keep this as-is for the RFC.

lutter: What I was getting at is that a `Chain` has a lifecycle, and rather than giving big prominence to the fact that it might want to have special things happening in the store, it might be better to have method stubs in this trait that calls out the fact that there is a lifecycle, and that the rest of the system will call certain things at certain points in the lifecycle. At a minimum, there will be a call when the chain is first created, which is when the chain would create custom schema, and there might well also be a need for a call when `graph-node` starts up and wants to get the chain ready for business.

leoyvens: I've added an `fn start()` method to the chain trait.

rfcs/0005-multi-blockchain-support.md
leoyvens: I've noted here that these components will need to be reworked to be made generic and reusable.

rfcs/0005-multi-blockchain-support.md
leoyvens: Thanks!

rfcs/0005-multi-blockchain-support.md
leoyvens: Added a note.

rfcs/0005-multi-blockchain-support.md
leoyvens: Updated it.

leoyvens:
Thanks for the reviews, I've addressed the comments.

